PR #18, opened 2015-01-10
only preload public posts; resolves #17

### LINKED ISSUE #17: Preload should only load public post types
Preload currently loads all posts with status `publish`. Not all of those will be publicly queriable, so on some sites there will be many requests that get 404'd. This is a waste of server resources. Better to just pull public post types, with the exception of attachment pages which also generally aren't required on the front end. PR coming.

webaware:
Resolves #17 

* fetch a list of post types that can be accessed from the front end;
* exclude `attachment` post type from the list;
* fetch IDs of posts in the remaining post types that have status `publish`

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -2966,7 +2966,11 @@ function wp_cron_preload_cache() {
 	}
 
 	if ( $wp_cache_preload_posts == 'all' || $c < $wp_cache_preload_posts ) {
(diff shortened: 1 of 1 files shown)


kraftbj:
I like this, except I don't want to exclude the attachment post type. Some sites do heavily use the attachment pages. Merging this in and then will commit to remove the unset for attachments.

webaware:
Fair enough, not something any of my sites use so it didn't seem important but I can see it has benefits.

Personally, I don't use preload on any sites. I erase the list of bots from Rejected User Agents and let the bots do the preload for me. They're all going to hit those posts anyway, so might as well benefit from that. But I don't have anything that presents different content to Google et al than to anonymous visitors.

kraftbj:
That's a nice way to handle it. Can't fault it :)

PR #21, opened 2015-01-15
Fixed - Fatal error: Internal Zend error - Missing class information

KoenRijpstra:
I had the same error as the following post: https://wordpress.org/support/topic/plugin-wp-super-cache-error-zend-on-gandi-hosting

Excluding the file from APC seems a bit to much trouble so i made this solution.

Can anybody confirm that this fixes the issue.

PHP: 5.3.10
APC: 3.1.7

### DIFF
--- a/wp-cache-base.php
+++ b/wp-cache-base.php
@@ -2,12 +2,7 @@
 $known_headers = array("Last-Modified", "Expires", "Content-Type", "Content-type", "X-Pingback", "ETag", "Cache-Control", "Pragma");
 
 if (!class_exists('CacheMeta')) {
-	class CacheMeta {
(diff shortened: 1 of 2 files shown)


kraftbj:
Cachemeta isn't actually being used anymore, so we're going to drop it totally via 81bbcfcf95d4c17825182b8d86804edc6a91e5f9

decvalts:
Thanks, this fixed the problems I had with APC errors caused by this class.

sunshinephotocart:
Has this been released yet? I am having this same issue but the current version doesn't seem to resolve it.

sccr410:
Has this been released yet? I have the most recent version and still having this issue.

PR #22, opened 2015-02-17
Cache JSON responses

ctimoteo:


ctimoteo:
I am starting using the WP REST API (http://wp-api.org/), 

i also want to use your plugin to cache the JSON responses from API, 

To permit caching i am ignoring "no closing HTML tags" (if JSON response)

The html comments that plugin produces also breaks the response. 

My patch ignores the wp_cache_add_to_buffer calls if JSON response.

the plugin works with my patch and it cache both JSON or html contents so i will share my patch anyway.

Thanks in advance.

wp-cache-phase1.php
```diff
@@ -506,7 +506,7 @@ function wp_cache_debug( $message, $level = 1 ) {
 		return false;
 
 	// Log message: Date URI Message
-	$log_messase = date('H:i:s') . " {$_SERVER['REQUEST_URI']} {$message}\n\r";
+	$log_message = date('H:i:s') . " {$_SERVER['REQUEST_URI']} {$message}\n\r";
```
kraftbj: Noting that this fix was made already in 0d392b07b71cbf21bb5ac71e6d52317f3fab2cc3

iandunn:
Another approach to this would be to [add regex patterns for JSON's object and collection syntax](https://wordpress.stackexchange.com/a/213331/3898).

Either way, I think it's important that the correct `content-type` header be set in `wp_cache_serve_cache_file()`. Sending JSON with a `text/html` header will cause the browser to render the page, which would be used as an XSS vector.

iandunn:
Regardless of how it's implemented, I think it's very important that WPSC adds support for caching the REST API, because the API is a big part of WP's future. The percentage of plugins and themes using it is increasing, and will continue to. 

The REST API is becoming a first-class citizen, and for those using it, caching its output is just as important as caching traditional pages. If WPSC doesn't support that, I think it'll end up being a nail in its coffin, as people will move on to other plugins that do.

It's also a pretty easy feature to add. Is there any practical reason not to?

MarkehMe:
Is this still happening? 

There is an alternative plugin here: https://en-au.wordpress.org/plugins/wp-rest-api-cache/

But it seems to not be caching API requests.

donnchawp:
This patch is very outdated now and unfortunately incomplete. It's not very efficient to json_decode() the buffer on every page load and there's no provision made for clearing parts of the REST API cache when the site is updated.

At the very least it should delete the entire REST API cache when an update occurs but preferably it should be more granular than that.

A more recent PR, Automattic/jetpack#25540 has a lot more going for it, and more discussion on the complexities of caching REST API responses.

PR #26, opened 2015-04-01
set url path for ossdl_off_cdn_url and ossdl_off_blog_url to home url

azanebrain:
Sets the URL path for CDN replacement to the front end URL of the site, instead of the backend.

This fixes an error on sites using a separate Site URL for the Administration Panels than the front end. For example, in [WordPress Skeleton](https://github.com/markjaquith/WordPress-Skeleton), the WordPress core files are put into the `/wp` directory, and can be accessed at http://mysite.com/wp. When the OSSDL off-linker goes to alter asset links to the off-site URL, it does not find any proper links. In this example, it is looking for mysite.com/**wp/**wp-content/file.jpg instead of mysite.com/wp-content/file.jpg.

### DIFF
--- a/ossdl-cdn.php
+++ b/ossdl-cdn.php
@@ -4,8 +4,8 @@
 
 /* Set up some defaults */
 if ( get_option( 'ossdl_off_cdn_url' ) == false )
-	add_option('ossdl_off_cdn_url', get_option('siteurl'));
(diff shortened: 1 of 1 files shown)


kraftbj:
Thanks for PR. We're review it shortly.

PR #37, opened 2015-06-02
Do not show WP Super Cache broken message for AJAX, REST, or XMLRPC requests

### LINKED ISSUE #36: wpcache_broken_message function breaks front-end AJAX
The advanced-cache.php file contains a function called wpcache_broken_message that runs on every front-end request (via register_shutdown_function) when there is an issue with the WP Super Cache setup and outputs this HTML comment...

`<!-- WP Super Cache is installed but broken. The constant WPCACHEHOME must be set in the file wp-config.php and point at the WP Super Cache plugin directory. -->`

The problem is, when you're trying to make a front-end AJAX request (instead of using admin-ajax.php) that expects a JSON response, that comment breaks the script.

I have confirmed that this issue also breaks JSON responses from the [WP REST API](https://wordpress.org/plugins/json-rest-api/).

Thanks!
Justin

kraftbj:
Fixes #36

### DIFF
--- a/advanced-cache.php
+++ b/advanced-cache.php
@@ -5,8 +5,14 @@ function wpcache_broken_message() {
 	if ( isset( $wp_cache_config_file ) == false )
 		return '';
 
-	if ( false == strpos( $_SERVER[ 'REQUEST_URI' ], 'wp-admin' ) )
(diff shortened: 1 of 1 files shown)


kraftbj:
Holding on merge to ensure it is complete wrt report in 36.

PR #40, opened 2015-06-23
'If-Modified-Since' isn't always defined?

mattclegg:
https://wordpress.org/support/topic/error-with-if-modified-since


kraftbj:
:+1: Thanks! Looks good to me.

Issue #33, opened 2015-05-15
Disabling cache flushes settings


kraftbj:
Unconfirmed via https://wordpress.org/support/topic/disabling-and-reenabling-the-cache-should-not-clear-all-my-settings?replies=1

jnylin:
I can confirm this. If I disable caching under the Easy-tab, the settings under Advanced is cleared, but only the first section.

Another strange behaviour, maybe related, is that if I change a setting in the first section under Advanced my settings about expiry time is cleared.

jnylin:
A relevant row: https://github.com/Automattic/wp-super-cache/blob/master/wp-cache.php#L450

jnylin:
And here is a relevant row for the other strange behaviour I descriped: https://github.com/Automattic/wp-super-cache/blob/master/wp-cache.php#L1055

jnylin:
I forked the repo so I when I get time I will try to fix it.

jnylin:
I created a pull request for the problem with cleared expiry time if you change settings. When I wrote this fix I also looked at the original problem with flushed settings and realised that it is that if you disable caching under the Easy tab it flushes all settings under the Advanced tab, if you disable caching under the Advanced tab the settings remain. I therefore think it's not a bug but a question about the admin interface.

kraftbj:
Solved via SHA:5987956 & #49

Issue #14, opened 2014-08-21
Expiry Time & Garbage Collection changing back to Clock schedule instead of Timer when submitting


glennsantos:
The fix was actually already implemented by someone, very fast fix.

Hope Automattic fixes this quickly. Can be frustrating for regular WP Super Cache users.

Fix here: http://wordpress.org/support/topic/found-problem-with-advanced-tab-garbage-collection-settings?replies=1

kraftbj:
Fixed via #49

PR #48, opened 2015-07-18
Don't clear expiry time if you change settings

jnylin:
Without this fix the settings for expiry time is cleared if you change any other setting under the Advanced tab, see issue #33


jnylin:
This will also avoid the PHP warning: ```Notice: Undefined index: cache_schedule_interval in .../wp-content/plugins/wp-super-cache/wp-cache.php on line 1624```

jnylin:
Seems like it's also related to #14

jnylin:
I close this, and do it in a separate branch.

PR #49, opened 2015-07-19
Don't flush settings for expiry time and garbage collection

jnylin:
This is a simple fix so that you can Update Status without clearing settings
for expiry time and garbage collection


jnylin:
This solves #14 and part of #33

kraftbj:
Thanks for the PR! I've been on the out of town/vacation/conference circuit so catching up now. I'm asking Donncha for a second set of eyes before pulling the trigger. Thanks!

PR #50, opened 2015-07-19
Check if listfile isset

jnylin:
This avoids a PHP notice if the user try to generate cache statistics and the cache is empty


kraftbj:
Thanks!

Issue #52, opened 2015-07-25
Blank page if wrong password


jnylin:
Not a big issue, but if cache is enabled and the user enters wrong password a blank page is returned.

This is what the log file shows for an unsuccessful login attempt:

    18:48:31 2585 /wordpress/wp-login.php In WP Cache Phase 2

    18:48:31 2585 /wordpress/wp-login.php Setting up WordPress actions

    18:48:31 2585 /wordpress/wp-login.php Created output buffer

    18:48:31 2585 /wordpress/wp-login.php supercache dir: /opt/lampp/apps/wordpress/htdocs/wp-content/cache/supercache/localhost/

    18:48:32 2585 /wordpress/wp-login.php Not caching POST request.

This is what the log file shows for a successful login attempt:

    18:50:35 2582 /wordpress/wp-login.php In WP Cache Phase 2

    18:50:35 2582 /wordpress/wp-login.php Setting up WordPress actions

    18:50:35 2582 /wordpress/wp-login.php Created output buffer

    18:50:35 2582 /wordpress/wp-login.php supercache dir: /opt/lampp/apps/wordpress/htdocs/wp-content/cache/supercache/localhost/

    18:50:35 2582 /wordpress/wp-login.php Not caching POST request.

    18:50:35 2582 /wordpress/wp-admin/ wp_cache_get_cookies_values: /^wp-postpass|^comment_author_|^wordpress_logged_in_/ Cookie detected: wordpress_logged_in_bbfa5b726c6b7a9cf3cda9370be3ee91

    18:50:35 2582 /wordpress/wp-admin/ supercache dir: /opt/lampp/apps/wordpress/htdocs/wp-content/cache/supercache/localhost/wordpress/wp-admin/

    18:50:35 2582 /wordpress/wp-admin/ No Super Cache file found for current URL: /opt/lampp/apps/wordpress/htdocs/wp-content/cache/supercache/localhost/wordpress/wp-admin/index.html

    18:50:35 2582 /wordpress/wp-admin/ wp_cache_get_cookies_values: cached: jakob|1438023035|b8P25ZKxCyfnelyWFJo7D4YnYTT3vtrIPxQace69qIM|0b2a7c5b5cd0b1c27e1ab0cdce06484d578e4aff0e9cacf2ad64d62aaa187624,

    18:50:35 2582 /wordpress/wp-admin/ In WP Cache Phase 2

    18:50:35 2582 /wordpress/wp-admin/ Setting up WordPress actions

[...]

PR #53, opened 2015-07-25
Fix blank page if cache is on and page shouldn't be cached

jnylin:
As I wrote in #52, the user was served a blank page if he/she entered wrong password at login. I also realized that if a page cache was on, the page shouldn't be cached, and there isn't a cache stored the user is served a blank page. 

This behavior was because wp_cache_maybe_dynamic was called with an undefined variable and this pull request fixes that.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -148,7 +148,7 @@ function wp_cache_get_response_headers() {
 function wp_cache_is_rejected($uri) {
 	global $cache_rejected_uri;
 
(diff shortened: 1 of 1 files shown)


jnylin:
Wrote in #52 how to reproduce the bug that this is a fix for.

jnylin:
Sorry, this fixes the blank page at login but not the blank page for rejected pages

jnylin:
This was solved by 9f8a03f7aefb494ae10db7ef1d55b20ff775bf39

PR #56, opened 2015-07-31
Import/Export settings

daneodekirk:
Adds an import/export feature as mentioned in #4. Provides the functionality for importing and exporting the plugin's settings as well as a class for validating/sanitizing the settings.


wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+  function __construct() {
+    global $wp_cache_config_file, $wp_cache_config_file_sample;
+
+    self::$cache_config_file = $wp_cache_config_file;
+    self::$cache_config_file_backup = str_replace( '.php', '-backup.php', self::$cache_config_file );
+
+    self::$MESSAGES = array(
+      0 =>  array( 'success', __( 'Settings imported', 'wp-super-cache' ) ),
+      1 =>  array( 'warning', __( 'Please upload an exported WP Super Cache settings file.', 'wp-super-cache' ) ),
+      2 =>  array( 'error', __( 'Unable to import the uploaded JSON file. Please export the settings and import them again.', 'wp-super-cache' ) ),
```
mdawaffe: I'd be annoyed if I got this error message. Why is it asking me to export something when what I'm trying to do is import? :)

mdawaffe:
Changelog: I think that, since import/export is a new feature, the changelog does not need to reflect all of the small changes and bug fixes.  @donnchawp may have other opinions.  Also, I don't think noting changlog changes in the changelog is important, though I see this has been done elsewhere in the plugin.

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+* by uploading an exported JSON file and overwriting the WP Super Cache configuration file with the
+* imported parameters. A backup of the original configuration file is kept.
+*
+* @package  WP Super Cache
+* @since 1.4.4
+*/
+
+class WP_Super_Cache_Export {
+
+  const TITLE = 'WP_Super_Cache_Export.json';
```
mdawaffe: I think of files as having names not titles. Are you the opposite?

mdawaffe:
There are some whitespace and semicolon inconsistencies.

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+   * An error message is displayed if the settings cannot be imported.
+   *
+   *  @since  1.4.4
+   */
+  static function form() {
+    ?>
+
+    <?php if (  isset( $_GET['message'] ) ):  ?>
+      <div id="message" class="notice notice-<?php echo self::$MESSAGES[ $_GET['message'] ][0] ; ?> is-dismissible">
+        <p><?php echo self::$MESSAGES[ $_GET['message'] ][1]  ?></p>
```
mdawaffe: We should escape output from translations.

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+      </div>
+    <?php endif; ?>
+
+      <fieldset class="options">
+        <h3><?php _e( "Export WP Super Cache Settings", "wp-super-cache" ) ?></h3>
```
mdawaffe: We should escape output from translation.  For this case and similar, we can use the `esc_html_e()` convenience function.

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+   */
+
+  public function import() {
+    if ( ! $this->can_import() ) {
+      return;
+    }
+
+
+    check_admin_referer( self::IMPORT_NONCE );
+    $file = $_FILES[ 'wp_super_cache_import_file' ][ 'tmp_name' ];
```
mdawaffe: We need to check `is_uploaded_file()`.

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+      exit;
+    }
+    $settings = (array) json_decode( file_get_contents( $file ), true );
+    if ( 0 === count( $settings ) ) {
+      wp_safe_redirect( add_query_arg( 'message', 2, $location ) );
+      exit;
+    }
+    if ( ! file_exists(self::$cache_config_file) )
+      wp_cache_verify_config_file();
+    // Backup the current config file to the same directory with a .backup file extension.
```
mdawaffe: This comment lies.

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+        exit;
+      }
+    }
+    $validator = new WP_Super_Cache_Sanitizer();
+    // // Update the database options that are not stored in the wp-cache-config.php file
+    if ( isset( $settings[ '_wp_super_cache_options' ] ) ) {
+      $options = array();
+      foreach ( $settings[ '_wp_super_cache_options'] as $key => $value ) {
+        $options[ $key ] = get_option( $key );
+        $validator->sanitize( $key, $value, $insert = true );
```
mdawaffe: What does this line do?

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+      return;
+    }
+    check_admin_referer(  self::EXPORT_NONCE );
+    include self::$cache_config_file;
+    $wp_cache_config_vars = get_defined_vars();
+    $wp_cache_config_options = $this->gather_plugin_options();
+    nocache_headers();
+    header( "Content-disposition: attachment; filename=" . self::TITLE );
+    header( 'Content-Type: application/octet-stream;' );
+    echo json_encode( array_merge($wp_cache_config_vars, $wp_cache_config_options )  );
```
mdawaffe: What do you think about pretty printing the JSON so that it's human readable?

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+   */
+
+  public function restore() {
+    if ( ! $this->can_restore() ) {
+      return;
+    }
+    check_admin_referer( self::RESTORE_NONCE );
+    $location = add_query_arg( 'tab', 'export', admin_url( 'options-general.php?page=wpsupercache' ) );
+    if ( file_exists(self::$cache_config_file) )
+      @unlink( self::$cache_config_file );
```
mdawaffe: Why is it important to unlink the file?

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+    if ( $this->backupFileExists() )
+      $renamed = @rename( self::$cache_config_file_backup, self::$cache_config_file );
+    if ( ! $renamed ) {
+      wp_safe_redirect( add_query_arg( 'message', 4, $location ) );
+      exit;
+    }
+    foreach ( get_option( '_wp_super_cache_backup_options', array() ) as $key => $value ) {
+        update_option( $key, $value );
+    }
```
mdawaffe: Do we need to call `schedule_wp_gc()` after a restore (like we do for import)?

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+    delete_option( '_wp_super_cache_backup_options' );
+    include self::$cache_config_file;
```
mdawaffe: Why is this include important?

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+
+  private function sanitize_value( $value, $setting ) {
+    if ( is_bool( $value ) ) {
+      return $value ? 1 : 0;
+    }
+    if ( is_numeric( $value ) ) {
+      return (int)  $value;
+    }
+    if ( is_string( $value ) ) {
+      $value = esc_html( strip_tags( $value ) );
```
mdawaffe: Are we using `esc_html()` just to make sure quotes don't get through?

wp-cache-export.php
```diff
@@ -0,0 +1,472 @@
...
+   * the html of string values.
+   *
+   * @since  1.4.4
+   *
+   * @uses esc_html Escape the string values
+   *
+   * @return number|string A sanitized version of the input value.
+   */
+
+  private function sanitize_value( $value, $setting ) {
```
mdawaffe: Where is this function called?

wp-cache-export.php
mdawaffe: Oh - I see.  The sanitizer also inserts.  That's not clear.

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+* this class's sanitize function.
+*
+* @package  WP Super Cache
+* @since 1.4.4
+*/
+
+class WP_Super_Cache_Sanitizer {
+
+  // The list of allowed database options
+  // Note: The stats and counts are commented out as they may not be necessary to import since they are site specific.
```
mdawaffe: We should make a decision.  If we're not going to import these, do we need to export them?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+   *
+   * @param String $setting The current setting of the value to be sanitized
+   * @param Mixed $value The value of the current setting being imported
+   * @param Boolean $insert Determines whether the value will be written to the wp-cache-config.php file or inserted into the database
+   *
+   * @since  1.4.4
+   */
+  function sanitize( $setting, $value, $insert = false ) {
+
+    if ( ! in_array( $setting, $this->allowed_values ) )
```
mdawaffe: We already have to check each array individually, so we may as well not create `$this->allowed_values` at all and do `else return false;` below.

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+   * are converted to strings to maintain the values in wp-cache-config.php file
+   *
+   * @param String $setting The current boolean setting to be sanitized
+   * @param Mixed $value The mixed value of the current setting that will be sanitized
+   *
+   * @return Boolean|String A sanitized version of the inputed value
+   */
+  function sanitize_boolean( $setting, $value ) {
+    // Note: necessary to force cache_enable and super_cache_enabled to be true/false
+    $sanitized_value = apply_filters( "wp_super_cache_sanitize_$value", $value );
```
mdawaffe: This filter doesn't work.

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+
+        case 'cache_acceptable_files' :
+            global $cache_acceptable_files;
+            if ( is_array( $value ) ) {
+                $value = implode( ',', $value );
+            }
+            $sanitized_value = wp_cache_sanitize_value( $value, $cache_acceptable_files );
+            break;
+
+        case 'cache_badbehaviour_file':
```
mdawaffe: This seems to never depend on the value we're importing.

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+   *
+   * @param String $setting The current integer setting to be sanitized
+   * @param Mixed $value The mixed value of the current setting that will be sanitized
+   *
+   * @return Integer A sanitized version of the inputed value
+   */
+  function sanitize_int( $setting, $value ) {
+    switch ( $setting ) {
+        case "cache_badbehaviour":
+            $value = $value == '' || $value == 'no' ? 0 : $value;
```
mdawaffe: Do we need this check? `intval( '' )` and `intval( 'no' )` are both `0`.

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+
+        case "cache_awaitingmoderation":
+        case "cache_wptouch":
+        case "cache_domain_mapping":
+            $value = $value == '' ? '0' : $value;
+            $sanitized_value = intval( $value );
+            break;
+
+        case "cache_time_interval":
+            $sanitized_value = 600;
```
mdawaffe: This is hard to grok.  The following would be much easer for me to read.

`$sanitized_value = is_numeric( $value ) ? intval( $value ) : 600;`

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+            break;
+
+        case "cache_max_time":
+            $sanitized_value = 3600;
+            $value = is_numeric( $value ) ? intval( $value ) : $sanitized_value;
+            if ( $sanitized_value !== $value )
+                $sanitized_value = $value;
+            break;
+
+        case "cache_rejected_uri":
```
mdawaffe: Why is this being sanitized with `::sanitize_int()`?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+  /**
+   * This sanitizes the settings where the allowed values are only integers.
+   * The sanitation method is different per setting name.
+   *
+   * @param String $setting The current integer setting to be sanitized
+   * @param Mixed $value The mixed value of the current setting that will be sanitized
+   *
+   * @return Integer A sanitized version of the inputed value
+   */
+  function sanitize_int( $setting, $value ) {
```
mdawaffe: For these integers, are we ever concerned about things like `0` or `-1000`?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+            if ( strlen( $value ) != 32 || ! ctype_xdigit( $value ) ) {
+                $sanitized_value = md5( date( 'H:i:s' ) . mt_rand() );
+            } else {
+                $sanitized_value = $value;
+            }
+            $sanitized_value = "\"$sanitized_value\"";
+            break;
+
+        case 'cached_direct_pages':
+            if ( is_array( $value ) )  {
```
mdawaffe: Why is this being sanitized with `::sanitize_string()`

mdawaffe:
Instead of manually constructing the strings to add to the config file (`"\"foo\""`, `'array(' . implode(...`), have you considered `var_export( $foo, true )`?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+                $sanitized_value = $value;
+            }
+            $sanitized_value = "\"$sanitized_value\"" ;
+            break;
+
+        case "wp_cache_home_path":
+            $sanitized_value = '/';
+            $home_path = parse_url( site_url() );
+            $home_path = trailingslashit( array_key_exists( 'path', $home_path ) ? $home_path[ 'path' ] : '' );
+            if ( isset( $wp_cache_home_path ) )
```
mdawaffe: When would this be true?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+                $sanitized_value =  $wp_cache_home_path;
+            if ( "$home_path" !== "$sanitized_value" )
+                $sanitized_value = $home_path;
+            $sanitized_value = "\"$sanitized_value\"" ;
+            break;
+
+        case 'wp_cache_mobile_browsers':
+            if ( function_exists( "cfmobi_default_browsers" ) ) {
+                $value = cfmobi_default_browsers( "mobile" );
+                $value = array_merge( $value, cfmobi_default_browsers( "touch" ) );
```
mdawaffe: Is it intentional that the imported value is only used if `cfmobi_default_browsers()` exists?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+                $value = self::MOBILE_BROWSERS;
+            }
+            $sanitized_value = apply_filters( 'cached_mobile_browsers', $value );
+            if ( is_array( $sanitized_value ) ) {
+                $sanitized_value = implode( ', ', $sanitized_value );
+            }
+            $sanitized_value = "\"$sanitized_value\"";
+            break;
+
+        case 'wp_cache_mobile_groups':
```
mdawaffe: This seems to never depend on the value we're importing.

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+            $value = apply_filters( 'cached_mobile_groups', array() );
+            if ( is_array( $value ) ) {
+                $sanitized_value = implode( ', ', $value );
+            } else {
+                $sanitized_value = '';
+            }
+            $sanitized_value = "\"$sanitized_value\"" ;
+            break;
+
+        // Note: this returns an array and not a string
```
mdawaffe: Then why is it here? :)

mdawaffe:
There are several settings in `WP_Super_Cache_Sanitizer` that always return their own thing no matter what value the user imported.  Should we even bother to output those settings?

wp-cache-sanitizer.php
```diff
@@ -0,0 +1,726 @@
...
+      "wp_cache_pages",
+      'wp_cache_plugins_dir',
+      "wp_cache_preload_email_volume",
+      "wp_cache_preload_posts",
+      "wp_super_cache_front_page_text",
+      "wptouch_exclude_ua",
+  );
+
+  // The default list of mobile browsers taken from wp-cache.php:648
+  const MOBILE_BROWSERS = array(
```
mdawaffe: Can we DRY this up and define them in only one place?

wp-cache-export.php
daneodekirk: I agree. I changed the constant here 63f39611.

wp-cache-export.php
daneodekirk: I actually felt the same way about that particular message. I made the message more specific 01489641.

wp-cache-export.php
daneodekirk: Good call. I replaced the translated text with their escaped html counterparts 802e94c5b.

wp-cache-export.php
daneodekirk: Replaced these as well 802e94c5b.

wp-cache-export.php
daneodekirk: Good catch. Added the check f40dec98.

wp-cache-export.php
daneodekirk: Yup. Edited to reflect the changed code  5adf7846.

wp-cache-export.php
daneodekirk: All good, I added a comment to clarify it anyways 81517e8

wp-cache-export.php
daneodekirk: It's a good option to have. I made the change to do so for PHP versions that have the option 65b6a1228c.

wp-cache-export.php
daneodekirk: It isn't, the rename will overwrite it anyways. I put it there initially for clarity. Removed b47f3280

wp-cache-export.php
daneodekirk: This was an oversight. Good catch 3c6fbad.

wp-cache-export.php
daneodekirk: It's not :). The idea was to explicitly load the setting variables before redirecting, but the scope makes that irrelevant 8370bf8d

wp-cache-export.php
daneodekirk: This was a legacy function from before the WP_Super_Cache_Sanitizer was coded. Good catch b842dbf90.

wp-cache-export.php
daneodekirk: Removed with the sanitize_value function  b842dbf90

wp-cache-sanitizer.php
daneodekirk: Good call. That's been implemented a294d2c470

wp-cache-sanitizer.php
daneodekirk: The wp_super_cache_sanitize_ filter has been replaced with using the setting name instead of the value c8e65e36

wp-cache-sanitizer.php
daneodekirk: I removed the check. It was originally there for clarity but intval does handle these values 9b1c23e0

wp-cache-sanitizer.php
daneodekirk: I implemented the new logic here and in another spot that followed the same code pattern 3c2a55c74

wp-cache-sanitizer.php
daneodekirk: This was a mistake. I moved it to the strings df20b7e0

wp-cache-sanitizer.php
daneodekirk: I did this with this setting and another one to explicitly show them being imported. Based on how the value is generated [here](https://github.com/Automattic/wp-super-cache/blob/master/plugins/badbehaviour.php#L57), I decided to import it but limit it to that value. This way someone can't edit the JSON file with a separate value for `cache_badbehaviour_file`.

wp-cache-sanitizer.php
daneodekirk: The settings that are arrays are outliers in the boolean, int, string paradigm. Since the arrays contain strings in them I decided to group them with the string settings. This is also the case for the other array settings and the `$wp_cache_pages` array as well. There is a case for regrouping these array settings separately.

wp-cache-sanitizer.php
daneodekirk: Good catch f3bb6cc5

wp-cache-sanitizer.php
daneodekirk: True. It's based on the logic [here](https://github.com/Automattic/wp-super-cache/blob/master/wp-cache.php#L685). Anything that filters the `cached_mobile_groups` should resolve the initial `$value`. Like the `cache_badbehaviour_file` it's here to protect against altered `cached_mobile_groups` settings.

wp-cache-sanitizer.php
daneodekirk: I ended up decoupling the array and string setting 4fb25f1

wp-cache-sanitizer.php
daneodekirk: The integer values are fairly restricted per setting. Tose miscellaneous numbers do not affect the logic when used in the plugin.

wp-cache-sanitizer.php
daneodekirk: We can, although we'll likely wait until after the exporter has been integrated to do so.

mdawaffe:
Closing this.  It'll be good inspiration for the next time we try to tackle this feature.

Issue #45, opened 2015-07-06
CPT with rewrite false clears complete cache


barrykooij:
If you've got a custom post type that has rewrite set to `false`, WP Super Cache clears the complete cache.

I believe this is happening because `get_current_url_supercache_dir` is stripping out the default WP rewrite and because of this the `$dir` because the root of the cache dir.

The content of `$permalink = trailingslashit( str_replace( get_option( 'home' ), '', post_permalink( $post_id ) ) );` (https://github.com/Automattic/wp-super-cache/blob/master/wp-cache-phase2.php#L1107) is `/?post_type=dlm_download&p=26/`. 

The content of `$dir = get_current_url_supercache_dir( $post_id );` is `/vagrant/download-monitor/wp-content/cache/supercache/lamp.dev/download-monitor/`. As you can see it completely strips of the permalink.

Since this CPT has no rewrite and no single I believe it shouldn't be cached at all. I think a check should be created checking if the result of `get_current_url_supercache_dir` isn't equal to the root cache dir.

barrykooij:
Any update on this?

kraftbj:
#66 will return out early if the two dirs match, which should resolve the immediate problem. I'd like to leave this open to audit where we can short-circuit the processes earlier for posts in this situation.

@barrykooij does #66 resolve it for you? Thanks for your patience!

PR #66, opened 2015-08-25
Do not clear the post cache if the current_url_supercache_dir matches…

kraftbj:
… the supercache_dir. 
See #45

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -1105,6 +1105,10 @@ function wp_cache_post_id_gc( $siteurl, $post_id, $all = 'all' ) {
 
 	$permalink = trailingslashit( str_replace( get_option( 'home' ), '', post_permalink( $post_id ) ) );
 	$dir = get_current_url_supercache_dir( $post_id );
+	if ( $dir == get_supercache_dir() ) {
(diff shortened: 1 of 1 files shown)


Sophist-UK:
For clarity, @graphicsxp was referring to #98 which he reported.

@graphicsxp - I think that this does fix the problem of all cache being deleted, but does NOT fix the problem of list pages (like blog pages showing all or specific categories of posts) needing to be invalidated. However this is a new problem which needs investigating.

graphicsxp:
ok then :) So it is not normal behaviour ?

Sophist-UK:
Actually,now I have reviewed the code again, I am not sure that this is the fix that is needed.

Function prune_super_cache is the routine which prunes the cache and it recursively prunes all subdirectories. This works for simple blogs where the category list has the posts in subdirectories, but is overkill for hierarchical pages (i.e. deletes too many pages), and can easily be incomplete i.e. it doesn't do archives, it doesn't do pages which e.g. use the list-category-posts plugin etc.

I think we should move this discussion to #98.

donnchawp:
A similar check was added in #420. Closing this.

PR #77, opened 2015-11-20
Make sure updated $cache_path ends in '/'

omegadrh:
`basename()` removes the trailing slash, which is expected in the `$cache_path` variable / config file. Without this fix, you can enter /some/path/wp-content/cache/ and end up with /some/path/wp-content/cachesupercache/ , .../wp-content/cachemeta/ , etc.

Also, don't bother updating the config file if the path hasn't changed.


kraftbj:
Looks good to me. Going to get a second :+1:.

PR #89, opened 2016-01-02
Fix lay-out on settings pages

tvdheijden:
Lay-out on settings page remains broken, even after https://github.com/Automattic/wp-super-cache/commit/ff1e165f1521e61176291b43c63a4c272f826f70. Grey line which should be below the tabs, is visible above the tabs.


donnchawp:
The changed by @Sophist-UK were also in another PR and need to be removed from this one.

PR #96, opened 2016-02-09
Defined some global vars used to prevent warnings with wp-cli

socketz:
Defined some global vars for prevent warning with wp-cli or other CLI scripts that uses wordpress. Probably exists other global vars to be defined.


socketz:
@stephdau @folletto @frosty @koke @rads Can you close if you don't like this solution?

PR #99, opened 2016-02-17
Fix front-page invalidation incorrect

Sophist-UK:
As far as I can see, this is deleting the front page cache when it is a static page rather than when it is a blog page.

This fix makes it delete the front page cache when the front page is a blog and not when it is a static page.


Sophist-UK:
Note: This fix is based only on reading the code. I have not tested that there is actually a problem with blog and static front pages nor tested that this fix corrects it.

donnchawp:
There was a different issue with that code which was fixed [here](https://github.com/Automattic/wp-super-cache/commit/98b536238669b023286c021118c9af22ebb6dee9).

PR #102, opened 2016-02-18
Do not clear existing cache on POST request

Sophist-UK:
A post request should definitely not be cached, but equally it should not invalidate a previous cache.


Sophist-UK:
Part 1 of 2 fixes to Automattic/wp-super-cache#959.

donnchawp:
I think this fixes the problem only because it stops the rebuild cleanup happening. The correct fix is in Automattic/wp-super-cache#186 which stops the rebuild system deleting files so much. Can you confirm?

There's now a check for POST requests in the phase1 script which disables the plugin early on.

PR #103, opened 2016-02-18
Avoid rejecting cache for root pages

Sophist-UK:
For home page $uri is either '' or '/' which will always be rejected.

So for these we will return false and leave it to other checks to avoid caching if needed.

Part 2 of 2 to address Automattic/wp-super-cache#959


donnchawp:
I couldn't replicate this. If the $uri is "/" then the code checks if any of the auto rejected strings are in "/" which isn't possible, it's not checking if "/" is in "/wp-admin/", it's checking if "/wp-admin/" is in "/".

PR #114, opened 2016-05-07
fix problem with wp-api

miya0001:
Related:
https://meta.trac.wordpress.org/ticket/1697

Thanks!


iandunn:
Related #22

kraftbj:
@miya0001 Could you rebase/force push this with only the functional changes so it's easier to read/review? I'll trim the whitespace separately to help keep the commits atomic. Thanks!

miya0001:
@kraftbj I'm done. 😄

donnchawp:
This isn't needed any more. The function wp_cache_add_to_buffer() [checks if $buffer contains "<html"](https://github.com/Automattic/wp-super-cache/blob/master/wp-cache-phase2.php#L542) and if it doesn't sends the output to the debug log, rather than the html comment.

PR #127, opened 2016-08-11
Remove `flush()` call in wp_cache_get_response_headers()

### LINKED ISSUE #126: Possible cause of `PHP Fatal Error:  Unknown: cannot use output buffering in output buffering display handlers in Unknown on line 0`
Setup: Apache 2.2.15 and php 5.5.38 (apache module) on CentOS (64 bit).

This error has been plaguing my logs for years (literally), while only happening sporadically, with no obvious patterns...

The trace that finally gave me something to go on was:

```
[client x.x.x.x] PHP Fatal error:  Unknown: Cannot use output buffering in output buffering display handlers in Unknown on line 0, referer: http://www.domain.tld
[client x.x.x.x] PHP Stack trace:, referer: http://www.domain.tld
[client x.x.x.x] PHP   1. shutdown_action_hook() /home/user/public_html/wp-includes/load.php:0, referer: http://www.domain.tld
[client x.x.x.x] PHP   2. do_action() /home/user/public_html/wp-includes/load.php:638, referer: http://www.domain.tld
[client x.x.x.x] PHP   3. call_user_func_array:{/home/user/public_html/wp-includes/plugin.php:525}() /home/user/public_html/wp-includes/plugin.php:525, referer: http://www.domain.tld
[client x.x.x.x] PHP   4. wp_ob_end_flush_all() /home/user/public_html/wp-includes/plugin.php:525, referer: http://www.domain.tld
[client x.x.x.x] PHP   5. ob_end_flush() /home/user/public_html/wp-includes/functions.php:3549, referer: http://www.domain.tld
[client x.x.x.x] PHP   6. wp_cache_ob_callback() /home/user/public_html/wp-includes/functions.php:3549, referer: http://www.domain.tld
[client x.x.x.x] PHP   7. wp_cache_shutdown_callback() /home/user/public_html/wp-content/plugins/wp-super-cache/wp-cache-phase2.php:314, referer: http://www.domain.tld
[client x.x.x.x] PHP   8. wp_cache_get_response_headers() /home/user/public_html/wp-content/plugins/wp-super-cache/wp-cache-phase2.php:867, referer: http://www.domain.tld
[client x.x.x.x] PHP   9. flush() /home/user/public_html/wp-content/plugins/wp-super-cache/wp-cache-phase2.php:124, referer: http://www.domain.tld
```

Which lead to:
https://github.com/Automattic/wp-super-cache/blob/3fce4ed10dd9a991197e917d2e9568c5516dd73a/wp-cache-phase2.php#L124

That line is there since forever:
https://github.com/Automattic/wp-super-cache/blame/3fce4ed10dd9a991197e917d2e9568c5516dd73a/wp-cache-phase2.php#L124
https://github.com/Automattic/wp-super-cache/commit/43acdbc4513fd23d5dbe3e875a84417ff7ba2b84 - commit log even references https://blog.saddey.net/2007/02/18/how-to-prevent-wpcache-from-changing-a-feeds-content-type-to-text-html-2/ ... 

So, that line was added to fix issues with missing content-type headers in feeds (and perhaps it did for that specific setup). But, it's still just blindly calling `flush()` while being god knows how many output buffering levels deep potentially? Not the safest thing to do really...

Anyway, just removing the `flush();` call completely should be perfectly safe, due to this:
https://github.com/Automattic/wp-super-cache/blob/3fce4ed10dd9a991197e917d2e9568c5516dd73a/wp-cache-phase2.php#L883

Removing it fixed the issue for me (not a single unknown fatal in the logs for over a week now, on 5 different sites of vastly different traffic patterns).

Quick PR to follow.

zytzagoo:
For more details see: https://github.com/Automattic/wp-super-cache/issues/126

Fixes #126

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -121,7 +121,6 @@ if ( !function_exists( 'wp_cache_user_agent_is_rejected' ) ) {
 
 function wp_cache_get_response_headers() {
 	if(function_exists('apache_response_headers')) {
-		flush();
 		$headers = apache_response_headers();
 	} else if(function_exists('headers_list')) {
 		$headers = array();


kraftbj:
At first read, your explanation and process makes sense to me. Donncha (primary dev) is on vacation and want to get his nod for modifying something this ancient ;-).

donnchawp:
We should check it in BUT we'll have to modify it because for some reason apache_response_headers() doesn't always behave and needs flush() to do the job. See [this comment](http://php.net/manual/en/function.apache-response-headers.php#56501) for more details.

I'll check the headers and if it's empty I'll call flush() and get the headers again.

DrLightman:
Buf `flush()` can't be used inside an output buffer display handler anyway, otherwise it triggers this PHP Fatal Error:

`PHP Fatal error:  Unknown: Cannot use output buffering in output buffering display handlers in Unknown on line 0`

Because `flush()` is used inside `wp_cache_get_response_headers()` that is used inside `wp_cache_shutdown_callback()` that is used inside `wp_cache_ob_callback()` that has been registered as an **output callback handler** with:

`ob_start( 'wp_cache_ob_callback' );`

donnchawp:
Ah, you're right. Take a look at #196 - I removed the flush() and changed the flow of the code so that headers_list() is used if apache_response_headers() fails.

PR #138, opened 2016-09-23
Do not cache/serve PUT or DELETE requests

### LINKED ISSUE #137: REST API: Do not cache PUT or DELETE requests
The WP REST API can use `PUT` and `DELETE` requests. We should treat these the same as we do a `POST` request and not cache them.

kraftbj:
Fixes #137


kraftbj:
WIP - Currently would prevent caching, but would still serve.

donnchawp:
Thumbs up from me! Looks good!

Issue #156, opened 2016-12-03
PHP 7.1 errors


sterndata:
After upgrading to PHP 7.1, I get this in my logs:

(this site 4.7 RC2)

`PHP Fatal error:  Uncaught Error: Cannot use string offset as an array in /var/www/stevenstern.us/wordpress/wp-content/plugins/wp-super-cache/wp-cache-phase2.php:584\nStack trace:\n#0 /var/www/stevenstern.us/wordpress/wp-content/plugins/wp-super-cache/wp-cache-phase2.php(313): wp_cache_get_ob('<!DOCTYPE html>...')\n#1 [internal function]: wp_cache_ob_callback('<!DOCTYPE html>...', 9)\n#2 /var/www/stevenstern.us/wordpress/wp-includes/functions.php(3673): ob_end_flush()\n#3 /var/www/stevenstern.us/wordpress/wp-includes/class-wp-hook.php(298): wp_ob_end_flush_all('')\n#4 /var/www/stevenstern.us/wordpress/wp-includes/class-wp-hook.php(323): WP_Hook->apply_filters('', Array)\n#5 /var/www/stevenstern.us/wordpress/wp-includes/plugin.php(453): WP_Hook->do_action(Array)\n#6 /var/www/stevenstern.us/wordpress/wp-includes/load.php(670): do_action('shutdown')\n#7 [internal function]: shutdown_action_!
 hook()\n#8 {main}\n  thrown in /var/www/stevenstern.us/wordpress/wp-content/plugins/wp-super-cache/wp-cache-phase2.php on line 584, referer: https://www.stevenstern.me/2015/01/stock-up-on-those-little-batteries-now-before/
`

and 

`legal string offset 'headers' in /var/www/stevenstern.us/wordpress/wp-content/plugins/wp-super-cache/wp-cache-phase2.php on line 584, referer: https://www.stevenstern.me/2015/01/pretty-damn-close-give-em-a-cigar-spacex/
`

and this on a 4.6 site:

[...]

PR #158, opened 2016-12-03
line 392 change to array() to avoid a PHP 7.1 error

sterndata:
This seems to fix the error I reported at https://github.com/Automattic/wp-super-cache/issues/156

Props to Otto for the code.


kraftbj:
This seems reasonable.

PR #161, opened 2016-12-08
Disable caching in customize preview

westonruter:
This is a first stab at ensuring compatibility with the customizer. 

A bug report for 4.7 seems to indicate it is not compatible: https://core.trac.wordpress.org/ticket/39128

I'm trying to pin down where the problem lies exactly, but here's an initial stab at compatibility.

### DIFF
--- a/wp-cache-phase1.php
+++ b/wp-cache-phase1.php
@@ -47,8 +47,9 @@ if ( isset( $wp_cache_make_known_anon ) && $wp_cache_make_known_anon )
 
 do_cacheaction( 'cache_init' );
 
-if (!$cache_enabled || $_SERVER["REQUEST_METHOD"] == 'POST')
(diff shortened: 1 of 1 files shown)


kraftbj:
@westonruter Thanks and this makes sense. Good to merge or do you think we should try to pin down things firmly first?

westonruter:
It does fix the one issue that I observed where changes failed to get applied in the preview after a refresh, so in that sense yes.

I've asked sixsigmablackbelt for more details on their configuration to figure out how to exactly reproduce the originally-reported issue.

westonruter:
See also https://core.trac.wordpress.org/ticket/39128#comment:27

donnchawp:
I can't replicate this problem at all. I'm using a multisite and domain mapping and customizer works ok but the fix above looks harmless so I'm happy to merge this.

Issue #169, opened 2017-01-27
On post update: too much stuff is deleted, not requested by me or settings?


DrLightman:
Hello, using WP 4.6.2, WPSC 1.4.8

I'm debugging on localhost, keeping an eye on supercache directory and debug log.

What I see is, when I update a post, all other 2 test posts in the blog (**ajsdjasj** and **qweqwe**) and the front page that I have in cache in that moment, get deleted too. I don't want this. The 2 posts make no sense, the front page per setting is off.

My garbage collection is set to 0 because I wanted it turned off while testing.

These are my settings:

mod_rewrite caching.

[x] Compress pages so they’re served more quickly to visitors. (Recommended)
[ ] Don’t cache pages for known users. (Recommended)
[ ] Don’t cache pages with GET parameters. (?x=y at the end of a url)
[x] Make known users anonymous so they’re served supercached static files.
[ ] Cache rebuild. Serve a supercache file to anonymous users while a new file is being generated. (Recommended)
[ ] Proudly tell the world your server is Stephen Fry proof! (places a message in your blog’s footer)

[x] Mobile device support. (External plugin or theme required. See the FAQ for further details.)

[ ] Remove UTF8/blog charset support from .htaccess file. Only necessary if you see odd characters or punctuation looks incorrect. Requires rewrite rules update.
[ ] Clear all cache files when a post or page is published or updated.
[ ] Extra homepage checks. (Very occasionally stops homepage caching) (Recommended)
[x] Only refresh current page when comments made.
[ ] List the newest cached pages on this page.
[ ] Coarse file locking. You probably don’t need this but it may help if your server is underpowered. Warning! May cause your server to lock up in very rare cases!
[ ] Late init. Display cached files after WordPress has loaded. Most useful in legacy mode.

$wp_cache_clear_on_post_edit = 0; //Added by WP-Cache Manager
in /wp-content/wp-cache-config.php

First 7 lines are ok. Then there are the **bolded** one that are the log lines that do the **extra unwanted pruning**, not requested by me. I'm going to edit post 1890 ( /video-singolo/ ) and I want only that cache to be pruned.

/wp-admin/post.php wp_cache_post_edit: Clearing cache for post 1890 on post edit.
/wp-admin/post.php supercache dir: D:\Siti\wptest\wp-content/cache/supercache/www.wptest.local/video-singolo/
[...]

PR #171, opened 2017-01-30
Ensure not to delete all cache on post update

DrLightman:
https://github.com/Automattic/wp-super-cache/issues/169

On post update `wp_cache_post_edit()` does the test on `$wp_cache_clear_on_post_edit` to decide to clear all cache or not, but when it calls `wp_cache_post_change()` this function can override that setting, so I added the test there too.

ps: sorry this is my first PR ever, I hope I got the procedure right.


donnchawp:
Thanks for the PR. It was the wrong code, but in the right area. Check out #175 which fixes the problem described in #169.

PR #186, opened 2017-02-06
Only define this rebuild directory if files are rebuilt.

donnchawp:
The plugin was deleting cache files too often because it thought that
they were rebuilt files. This patch waits until files are rebuilt to
define that rebuild list.


webtaster:
Thanks, it fixed the issue for my site.

PR #188, opened 2017-02-06
Apply realpath to $cache_path to avoid problems on Windows

### LINKED ISSUE #183: realpath() causing get_all_supercache_filenames() to return an empty array in some cases
I think this mostly affects **Windows** systems. Fortunately the fix shoud be **very easy**.

How did I notice?

While debugging the new 1.4.9 for some strange behaviours, in the code below `$files_to_check` was empty:

(wp-cache-phase2.php, line 1201)
```php
	if ( $all == true ) {
		wp_cache_debug( "Post change: supercache enabled: deleting cache files in " . $cache_path . 'supercache/' . $siteurl, 4 );
		$files_to_check = get_all_supercache_filenames( $dir );
```

The code of interest inside `get_all_supercache_filenames()`:

(wp-cache-phase1.php, line 618)
```php
function get_all_supercache_filenames( $dir = '' ) {
	global $wp_cache_mobile_enabled, $cache_path;

	$dir = realpath( $dir );
	if ( substr( $dir, 0, strlen( $cache_path ) ) != $cache_path )
		return array();
```

The problem there is that in **Windows** systems, like the one I am developing on, `realpath()` among the other things of espanding .. or trimming the ending slash, also converts _forward slashes_ `/` in _backslashes_ `\`:

`$dir`
D:\Siti\wptest\wp-content/cache/supercache/www.wptest.local/

`realpath( $dir )`
D:\Siti\wptest\wp-content\cache\supercache\www.wptest.local

So you're already seeing where this is going. The test in the IF `substr( $dir, 0, strlen( $cache_path ) ) != $cache_path` will succeed when it shouldn't just becase there are no forward slashes in the left argument whereas `$cache_path` in the right argument has not undergone the same conversion and could contain some forward slashes:

`substr( $dir, 0, strlen( $cache_path ) )`
D:\Siti\wptest\wp-content\cache\

`$cache_path`
D:\Siti\wptest\wp-content/cache/

So these two are the same directory alright but the != test is flawed.

**The fix I propose here seems very easy and consists in applying realpath to `$cache_path` too, before the IF test, maybe using a local variable `$cache_path_rp` for not overwriting the global `$cache_path`**:

(wp-cache-phase1.php, line 618)
```php
function get_all_supercache_filenames( $dir = '' ) {
	global $wp_cache_mobile_enabled, $cache_path;

	$dir = realpath( $dir );
	$cache_path_rp = realpath( $cache_path ); // NEW
	if ( substr( $dir, 0, strlen( $cache_path_rp ) ) != $cache_path_rp )
		return array();
```

I rapidly checked all the plugin for other possibile bugs caused by `realpath()`, but I don't think there are others, but maybe I'm wrong.

donnchawp:
Fixes #183.
realpath() on Windows will change a UNIX style path to a Windows C:\
type path.

### DIFF
--- a/wp-cache-phase1.php
+++ b/wp-cache-phase1.php
@@ -721,10 +721,10 @@ function wp_cache_confirm_delete( $dir ) {
 	// don't allow cache_path, blog cache dir, blog meta dir, supercache.
 	$dir = realpath( $dir );
 	if ( 
-		$dir == $cache_path || 
-		$dir == $blog_cache_dir ||
-		$dir == $blog_cache_dir . "meta/" ||
-		$dir == $cache_path . "supercache"
+		$dir == realpath( $cache_path ) ||
+		$dir == realpath( $blog_cache_dir ) ||
+		$dir == realpath( $blog_cache_dir . "meta/" ) ||
+		$dir == realpath( $cache_path . "supercache" )
 	) {
 		return false;
 	} else {
--- a/wp-cache-phase1.php
+++ b/wp-cache-phase1.php
@@ -619,7 +619,8 @@ function get_all_supercache_filenames( $dir = '' ) {
 	global $wp_cache_mobile_enabled, $cache_path;
 
 	$dir = realpath( $dir );
-	if ( substr( $dir, 0, strlen( $cache_path ) ) != $cache_path )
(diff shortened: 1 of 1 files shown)


donnchawp:
Thank you for spotting that!

PR #250, opened 2017-06-02
Added notices for preload page

donnchawp:


rest/class.wp-super-cache-rest-get-notices.php
```diff
@@ -28,6 +29,40 @@ public function callback( $request ) {
...
+		if ( false == $super_cache_enabled ) {
+			$notices[ 'preload_disabled_supercache_off' ] = array(
+				'type' => 'warning',
+				'message' => __(
+					'Preloading is disabled as supercaching is disabled',
+					'wp-super-cache'
+				),
+			);
+		}
+		if ( false === defined( 'DISABLESUPERCACHEPRELOADING' ) ) {
```
donnapep: @donnchawp Should this be checking for `true === defined( 'DISABLESUPERCACHEPRELOADING' )`?

rest/class.wp-super-cache-rest-get-notices.php
```diff
@@ -28,6 +29,40 @@ public function callback( $request ) {
...
+		if ( false == $cache_enabled ) {
+			$notices[ 'preload_disabled_cache_off' ] = array(
+				'type' => 'warning',
+				'message' => __(
+					'Preloading is disabled as caching is disabled',
+					'wp-super-cache'
+				),
+			);
+		}
+		if ( false == $super_cache_enabled ) {
```
donnapep: @donnchawp How do I enable supercache in order to test w/o this notice? This line is already in `wp-cache-config.php`:

`$super_cache_enabled = true;`

rest/class.wp-super-cache-rest-get-notices.php
donnchawp: You're right! Good catch!

rest/class.wp-super-cache-rest-get-notices.php
donnchawp: If you enable legacy mode caching that will disable super cache mode, and set `$super_cache_enabled = false;`

donnapep: I can get it to show, but I can't get it to go away. 😄 Switching to PHP or mod_rewrite still returns the `preload_disabled_supercache_off` notice.

Issue #224, opened 2017-04-10
Description of Caching Types


donnchawp:
I want to change the descriptions and even names of the different caching types to make it easier for new users to understand what settings they should use. At present we have:

1. mod_rewrite
2. PHP
3. Legacy page caching

This exposes the implementation of the caching delivery, not what they're useful for. A technical user might understand the implications of the first two but not the third. Any other user will be lost but at least the settings page recommends PHP mode.

Related to legacy caching is the option "Don’t cache pages for known users." which is recommended. This disables a portion of legacy page caching. That stops caching when logged in users or users who comment visit the site, except for feeds because of [this code](https://plugins.trac.wordpress.org/changeset/331205).

There's also "Don’t cache pages with GET parameters" which disables the other function of legacy caching which is to cache pages that have ?x=y or whatever at the end of the URL.

Explanation:
1. mod_rewrite mode uses rules in a .htaccess file to serve cached files instead of WordPress loading, hitting the database and taking time. It can be blazingly fast and a server will be able to cope well under load as long as the web server is configured correctly. Writing to the .htaccess file of a live website can be risky but the code that does this has been working fairly reliably for many years now.
2. PHP mode uses PHP to serve cache files in a similar way. A portion of WordPress is loaded, and since WordPress 4.6 more of the system is loaded but it's still very light on resources. This is recommended because it's simple to get going.
3. Legacy Caching is the method WP-Cache used to store cached pages. These cache files store information about the headers and are stored as an md5 hash of the url and identifying user information. This means that we can cache pages with personalised content such as name or username. Previously the cache files were stored in one directory. Maintaining them was relatively expensive because garbage collection had to scan all files to find the files related to the current page (for example). The next release of the plugin will store these files in the supercache directories make maintenance simple. 

Methods 1 and 2 are used only to store hits from anonymous, unknown users. Legacy caching is always used (except when the options above are selected) to cache files for known users like those that are logged in or leave comments.

With the changes to legacy caching I feel more confident about people using it more. It should be used to cache REST API calls for example.

So what should we call them?

Maybe separate them out to "anonymous users" and "full caching" groups? 

One thing to consider is that legacy caching always works anyway unless deliberately disabled so maybe we should only ask the user to choose between the first two caching methods, and offer options to disable caching for known users and the GET parameters..

marcomarsala:
1) Full page cache (static)
2) Full page cache (dynamic)
I'll remove ability to select 3).

[...]

PR #255, opened 2017-06-07
Update settings page cache on/off process.

donnchawp:
See #224.

Legacy caching wasn't a great name for the "default caching" this plugin
does, and only technical people will recognise PHP and mod_rewrite so we
simplified it.

Simple = PHP
Expert = mod_rewrite

And no mention of legacy caching which has become "standard caching".


donnchawp:
A weakness of this UI is that "Standard Caching" has to be enabled for "Super Caching" to be enabled. The control for both maybe should be one element. Maybe radio boxes?
1. Standard Caching
2. Super Caching (includes Standard Caching)

donnchawp:
After the last code check in:

screen shot 2017-06-08 at 13 17 58

donnchawp:
And now it's even simpler:

screen shot 2017-06-08 at 16 58 29

PR #258, opened 2017-06-14
Update the .htaccess rules safely without bringing down the site

donnchawp:
mod_rewrite rules have to be updated to use Export mode but if done
incorrectly the site can be taken offline. This code makes a backup of
the .htaccess file, grabs a copy of the front page, updates the rules,
grabs the front page again and if they are not the same then it restores
the backup.
TODO: more error reporting as there are so many checks.


donnchawp:
Works for multisite installs using domains and directories. I'm going to change the wsc_mod_rewrite() function so it uses the new update function and get rid of some of the warnings that rules can't be updated.

PR #264, opened 2017-06-20
Change "notice(s)" to status

donnchawp:
Renaming notices to status.


rest/class.wp-super-cache-rest-get-status.php
```diff
@@ -0,0 +1,114 @@
...
+
+	/**
+	 * @param array $status
+	 */
+	protected function add_compression_status( & $status ) {
+		if ( defined( 'WPSC_DISABLE_COMPRESSION' ) ) {
+			$status['compression_disabled'] = true;
+		} elseif ( false == function_exists( 'gzencode' ) ) {
+			$status['compression_disabled'] = true;
+		}
```
ockham: Hmm, the downside here is that we lose information -- the client can't give error notices at the same level of granularity [as before](https://github.com/Automattic/wp-super-cache/pull/264/files#diff-e44368ece331b604e3ad68985462b07bL108).

Wonder if we should change the status from bool to a string const that indicates the reason, or split into two separate stati?

donnchawp: I think we'll need new status endpoints where the REST API client can request more information. Not every client is going to care about why something is disabled, probably.

donnchawp: For completeness, fixed this in #271 by splitting the "compression disabled" status into two.

PR #311, opened 2017-07-24
Load the REST API using dirname( __FILE__ ), it's safer.

donnchawp:
Sometimes WPCACHEHOME isn't reliable.


donnchawp:
The plugin doesn't check WPCACHEHOME enough, as explained in [this forum post](https://wordpress.org/support/topic/2-days-and-still-no-fix-any-eta/#post-9341361).

PR #322, opened 2017-07-26
Fix a critical bug that deletes wp-admin when deleting the plugin

### LINKED ISSUE #321: Critical Bug - Deactivate and Delete Blows Away wp-admin
Greetings, there is a critical bug in the latest release of this plugin, Deactivating the plugin and then Deleting it within wp-admin in a deactivated state will cause the wp-admin folder to be fully unlinked and removed.

I believe it is related to these commits

https://github.com/Automattic/wp-super-cache/commit/1b34fbd2068296300967723ec84aec0f39fb2835

https://github.com/Automattic/wp-super-cache/commit/57ac2773c58dc8919bbc7df2f3bf5c4824914234

when deactivated WPCACHEHOME no longer exists in the wp-config.php file, and since you are within /wp-admin/ I 'think' it is deciding that wp-admin is now WPCACHEHOME and as such unlinking the entire directory in the process of cleaning up. I've done multiple installs and had another independent wordpress dev confirm as well, assistance is appreciated.

dbtlr:
Fix a critical bug that deletes wp-admin when deleting the deactivated wp-super-cache plugin

This fixes #321 by making sure that unlink is not run on NULL or empty cache directories that appear to be leaking into this from some other source.

### How to test

1. Checkout branch
2. Activate WP Super Cache (from plugins manager)
3. Deactivate WP Super Cache (from plugins manager)
4. Delete WP Super Cache (from plugins manager)
5. Verify wp-admin was not deleted.

### Future considerations

We should look to adding some sort of tracking to by wrapping the function in a `wpsc_unlink` function that not only verifies that files are valid but also allows for some sort of optional logging for debugging these sorts of issues, to see what is deleted and from where.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -874,6 +874,12 @@ function wp_cache_phase2_clean_cache($file_prefix) {
 }
 
 function prune_super_cache( $directory, $force = false, $rename = false ) {
+
+	// Don't prune a NULL/empty directory.
+	if ( null === $directory || '' === $directory ) {
+		return false;
+	}
+
 	global $cache_max_time, $cache_path, $blog_cache_dir;
 	static $log = 0;
 	static $protected_directories = '';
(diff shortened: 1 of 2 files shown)


juraisa:
Thanks for the quick work all

juraisa:
Thanks all, be sure to ping WP so they can reopen the plugin on their end

PR #332, opened 2017-07-31
Add "site url" setting to CDN page to avoid problems with that setting

donnchawp:
WordPress installs "in their own directory" will sometimes have
different siteurl and home url which can cause problems. This setting
will allow the site administrator to set it correctly.

### DIFF
--- a/ossdl-cdn.php
+++ b/ossdl-cdn.php
@@ -5,7 +5,9 @@
 /* Set up some defaults */
 if ( get_option( 'ossdl_off_cdn_url' ) == false )
 	add_option( 'ossdl_off_cdn_url', get_option( 'siteurl' ) );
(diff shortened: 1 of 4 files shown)


ossdl-cdn.php
```diff
@@ -179,6 +182,13 @@ function scossdl_off_options() {
 				<th scope="row"><label for="ossdlcdn"><?php _e( 'Enable CDN Support', 'wp-super-cache' ); ?></label></th>
 			</tr>
 			<tr valign="top">
+				<th scope="row"><label for="ossdl_off_cdn_url"><?php _e( 'Site URL', 'wp-super-cache' ); ?></label></th>
```
ockham: `for=` should be `ossdl_off_blog_url`

PR #333, opened 2017-08-01
Fix PHP Coding Standards errors

Ismail-elkorchi:
Reference: https://make.wordpress.org/core/handbook/best-practices/coding-standards/php/


donnchawp:
Wow. This is quite a bit of code. Thank you. The best way to submit something like this would be in smaller chunks however. Reviewing one huge change like this is difficult, but I'll have a look.

Ismail-elkorchi:
@donnchawp My pleasure.

> The best way to submit something like this would be in smaller chunks however.

Honestly, I find it very difficult to submit changes in smaller chunks, since there are a huge number of WPCS violations in the plugin. For example, wp-cache.php has more than 3700 violations, wp-cache-phase2.php has more than 1000 violations, wp-cache-phase2.php more than 500, and so on.

What is the best way to handle this change ? Should I deal with each file in a new PR ? Or should I create one PR and submit in each commit the changes made to one file ?

These two solutions would have the same drawback that you stated in your response, and I don't see how should I submit the remaining fixes in a manner that makes the review easier for you?

What do you see as the best way to organize this work?

donnchawp:
There's no need to resubmit these changes. I looked over most of the changes but by the time I got to wp-cache.php I was square-eyed from looking at green/red lines in the changed code.

I'll download these files and merge them individually, probably bunching the plugins into one PR, and the major PHP files into separate PRs, but it might not be for a while because I want to get a small release out this week.

Ismail-elkorchi:
@donnchawp Thank you for your review. I adjusted the code based on your feedback.

I'll create the new PRs as you suggested once this PR is merged.

PR #336, opened 2017-08-02
Better notifications on preload page and revert sql to fetch posts

### LINKED ISSUE #326: Preload issue after 1.5.0 - wrong counting posts
It's related to commit [Added post_count setting to be used by preload UI ](https://github.com/Automattic/wp-super-cache/commit/2b3a4779030187bbd9634f8e82a45eabc3e091b0)

Test of function [wp_count_posts](https://codex.wordpress.org/Function_Reference/wp_count_posts):
```php
$count = wp_count_posts();
echo $count->publish, PHP_EOL;

$count = wp_count_posts( 'post' );
echo $count->publish, PHP_EOL;

$count = wp_count_posts( 'page' );
echo $count->publish, PHP_EOL;

$count = wp_count_posts( 'product' );
echo $count->publish, PHP_EOL;
```
Output:
9
9
39
1053

## Usage

`<?php wp_count_posts( $type, $perm ); ?>`

## Parameters

**$type**
(string) (optional) Post type to count
Default: 'post' 

## SQL query
`SELECT post_status, COUNT( * ) AS num_posts FROM wp_posts WHERE post_type = 'post' GROUP BY post_status`

## Consequence:

```
15:27:45 28821 /wp-cron.php wp_cron_preload_cache: no more posts to get. Limit (9) reached.
15:27:45 28821 /wp-cron.php wp_cron_preload_cache: no more posts. scheduling next preload in 720 minutes.
```
Only first of 100 posts/pages/products are fetched.

Possible solution is usage of direct queries again (as in version 1.4.9).

donnchawp:
Fixes #326

The function wp_count_posts() only fetches posts of type "post" and some
users have other types of posts to preload.
The preload page notifications weren't clear enough, so this makes them
into proper admin notices.

### DIFF
--- a/rest/class.wp-super-cache-rest-get-settings.php
+++ b/rest/class.wp-super-cache-rest-get-settings.php
@@ -172,7 +172,7 @@ class WP_Super_Cache_Rest_Get_Settings extends WP_REST_Controller {
 	 */
 	protected function get_minimum_preload_interval() {
 		global $wpdb;
-		$count = wpsc_post_count();
+		$count = $this->get_post_count();
 		if ( $count > 1000 ) {
 			$min_refresh_interval = 720;
 		} else {
@@ -197,8 +197,7 @@ class WP_Super_Cache_Rest_Get_Settings extends WP_REST_Controller {
 	 * @return int
 	 */
 	protected function get_post_count() {
-		$count = wpsc_post_count();
-		return $posts_count->publish;
+		return wpsc_post_count();
 	}
 
 	/**
--- a/rest/class.wp-super-cache-rest-get-settings.php
+++ b/rest/class.wp-super-cache-rest-get-settings.php
@@ -172,8 +172,7 @@ class WP_Super_Cache_Rest_Get_Settings extends WP_REST_Controller {
 	 */
 	protected function get_minimum_preload_interval() {
 		global $wpdb;
-		$posts_count = wp_count_posts();
-		$count = $posts_count->publish;
+		$count = wpsc_post_count();
 		if ( $count > 1000 ) {
 			$min_refresh_interval = 720;
 		} else {
@@ -198,7 +197,7 @@ class WP_Super_Cache_Rest_Get_Settings extends WP_REST_Controller {
 	 * @return int
 	 */
 	protected function get_post_count() {
-		$posts_count = wp_count_posts();
+		$count = wpsc_post_count();
 		return $posts_count->publish;
 	}
 
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -821,6 +821,45 @@ table.wpsc-settings-table {
 		$_GET[ 'tab' ] = 'easy';
 	}
 
+	if ( $_GET[ 'tab' ] == 'preload' ) {
+		if ( $super_cache_enabled == true && false == defined( 'DISABLESUPERCACHEPRELOADING' ) ) {
+			global $wp_cache_preload_interval, $wp_cache_preload_on, $wp_cache_preload_taxonomies, $wp_cache_preload_email_me, $wp_cache_preload_email_volume, $wp_cache_preload_posts, $wpdb;
+			$count = wpsc_post_count();
+			if ( $count > 1000 ) {
+				$min_refresh_interval = 720;
+			} else {
+				$min_refresh_interval = 30;
+			}
+			$return = wpsc_preload_settings( $min_refresh_interval );
+			$msg = '';
+			if ( empty( $return ) == false ) {
+				foreach( $return as $message ) {
+					$msg .= $message;
+				}
+			}
+			$currently_preloading = false;
+
+			$preload_counter = get_option( 'preload_cache_counter' );
+			if ( isset( $preload_counter[ 'first' ] ) ) // converted from int to array
+				update_option( 'preload_cache_counter', array( 'c' => $preload_counter[ 'c' ], 't' => time() ) );
+			if ( is_array( $preload_counter ) && $preload_counter[ 'c' ] > 0 ) {
+				$msg .= '<p>' . sprintf( __( 'Currently caching from post %d to %d.', 'wp-super-cache' ), ( $preload_counter[ 'c' ] - 100 ), $preload_counter[ 'c' ] ) . '</p>';
+				$currently_preloading = true;
+				if ( @file_exists( $cache_path . "preload_permalink.txt" ) ) {
+					$url = file_get_contents( $cache_path . "preload_permalink.txt" );
+					$msg .="<p>" . sprintf( __( "<strong>Page last cached:</strong> %s", 'wp-super-cache' ), $url ) . "</p>";
+				}
+				if ( $msg != '' ) {
+					echo '<div class="notice notice-warning"><h3>' . __( 'Preload Active', 'wp-super-cache' ) . '</h3>' . $msg . '</div>';
+				}
+			}
+			next_preload_message( 'wp_cache_preload_hook', __( 'Refresh of cache in %d hours %d minutes and %d seconds.', 'wp-super-cache' ), 60 );
+			next_preload_message( 'wp_cache_full_preload_hook', __( 'Full refresh of cache in %d hours %d minutes and %d seconds.', 'wp-super-cache' ) );
+
+		}
+	}
+
+
 	wpsc_admin_tabs();
 
 	if ( isset( $wp_super_cache_front_page_check ) && $wp_super_cache_front_page_check == 1 && !wp_next_scheduled( 'wp_cache_check_site_hook' ) ) {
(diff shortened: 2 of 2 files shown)


rest/class.wp-super-cache-rest-get-settings.php
```diff
@@ -198,7 +197,7 @@ protected function get_is_lock_down_enabled() {
 	 * @return int
 	 */
 	protected function get_post_count() {
-		$posts_count = wp_count_posts();
+		$count = wpsc_post_count();
```
stodorovic: It's wrong. `wpsc_post_count` return integer instead array(). It should be:
```php
        protected function get_post_count() {
                return wpsc_post_count();
        }
```

stodorovic: Maybe, we can use `$count = $this->get_post_count();` instead `$count = wpsc_post_count();`

stodorovic:
I'm testing this PR. I'll send better feedback tomorrow.

rest/class.wp-super-cache-rest-get-settings.php
```diff
@@ -172,8 +172,7 @@ protected function get_is_preload_active() {
 	 */
 	protected function get_minimum_preload_interval() {
 		global $wpdb;
-		$posts_count = wp_count_posts();
-		$count = $posts_count->publish;
+		$count = wpsc_post_count();
```
stodorovic: Maybe, we can use `$count = $this->get_post_count();` instead `$count = wpsc_post_count();` It seems as more flexible solution.

rest/class.wp-super-cache-rest-get-settings.php
donnchawp: Thanks, well spotted!

stodorovic:
I've done few tests on various websites. There are some CPTs which aren't public (or don't have rewrite rules), but they have set _post_status=public_. Consequence of this is inconsistency (in number of posts) between function `wpsc_post_count` and _query_ which gets post IDs. ( I didn't do detail checking before).

Simple and more efficient solution is to add similar filter as `wp_cache_preload_taxonomies` instead of filter `wpsc_post_count_sql`. This filter could be used in both queries.

Proposal for function `wpsc_post_count`:

```php
function wpsc_post_count() {
        global $wpdb;
        static $count;
        
        if ( isset( $count ) ) {
                return $count;
        }
        
        $post_types = (array) apply_filters( 'wp_cache_preload_post_types', get_post_types(
                array( 
                        'public' => true, 
                        'publicly_queryable' => true
                ),
                'names', 
                'or' )
        );
        
        $count = $wpdb->get_var( "SELECT COUNT(*) FROM {$wpdb->posts} WHERE post_type IN ('" .
                join( "', '", array_map( 'esc_sql', $post_types ) ) .
                "') AND post_status = 'publish'" );

        return $count;
}
```

I didn't create PR because I'm not sure which is right coding style here. Static cache doesn't have effects because I see that previous filter is triggered only once per HTTP request( only from _page=wpsupercache&tab=preload_).
So I've reduced complexity because it doesn't make sense to someone changes post_types more than once per HTTP request. Also, I think that isn't good to leave option to overwrite entire SQL query.

We should add same filter for query in `wp-cache.php`. Proposal for the code:
```php
        $post_types = (array) apply_filters( 'wp_cache_preload_post_types', get_post_types(
                array( 
                        'public' => true, 
                        'publicly_queryable' => true
                ),
                'names', 
                'or' )
        );
      
        $posts = $wpdb->get_col( "SELECT ID FROM {$wpdb->posts} WHERE post_type IN ('" .
                join( "', '", array_map( 'esc_sql', $post_types ) ) .
                "') AND post_status = 'publish' ORDER BY ID DESC LIMIT $c, 100" );
```

Maybe it's good to create function which gets post_types, but it's only two places in the code for now. I don't know what's better.

I'll do detail test again (on all websites which I maintains) when we merge this PR.

stodorovic:
Couple tests:

```php
add_filter( 'wp_cache_preload_post_types', '__return_false' );
echo wpsc_post_count2(), PHP_EOL;

add_filter( 'wp_cache_preload_post_types', '__return_empty_array' );
echo wpsc_post_count2(), PHP_EOL;

add_filter( 'wp_cache_preload_post_types', function(){ return 'post'; } );
echo wpsc_post_count2(), PHP_EOL;

add_filter( 'wp_cache_preload_post_types', function(){ return array( 'post', 'page' ); } );
echo wpsc_post_count2(), PHP_EOL;
```
_wpsc_post_count2 doesn't have cache because I need to run it few times for testing purpose_

SQL Queries:
```php
    [17] => Array
        (
            [0] => SELECT COUNT(*) FROM wp_posts WHERE post_type IN ('') AND post_status = 'publish'
            [1] => 5.1975250244141E-5
            [2] => wpsc_post_count2
        )

    [18] => Array
        (
            [0] => SELECT COUNT(*) FROM wp_posts WHERE post_type IN ('') AND post_status = 'publish'
            [1] => 4.9114227294922E-5
            [2] => wpsc_post_count2
        )

    [19] => Array
        (
            [0] => SELECT COUNT(*) FROM wp_posts WHERE post_type IN ('post') AND post_status = 'publish'
            [1] => 5.0067901611328E-5
            [2] => wpsc_post_count2
        )

    [20] => Array
        (
            [0] => SELECT COUNT(*) FROM wp_posts WHERE post_type IN ('post', 'page') AND post_status = 'publish'
            [1] => 6.3180923461914E-5
            [2] => wpsc_post_count2
```
Output:
0
0
6
45

So, filter covers also stupid actions without errors.

```
mysql> explain SELECT COUNT(*) FROM wp_posts WHERE post_type IN ('post', 'page') AND post_status = 'publish';
+----+-------------+----------+-------+------------------+------------------+---------+------+------+--------------------------+
| id | select_type | table    | type  | possible_keys    | key              | key_len | ref  | rows | Extra                    |
+----+-------------+----------+-------+------------------+------------------+---------+------+------+--------------------------+
|  1 | SIMPLE      | wp_posts | range | type_status_date | type_status_date | 124     | NULL |   77 | Using where; Using index |
+----+-------------+----------+-------+------------------+------------------+---------+------+------+--------------------------+
1 row in set (0.00 sec)

mysql> explain SELECT COUNT(*) FROM wp_posts WHERE post_status = 'publish';                                  
+----+-------------+----------+-------+---------------+------------------+---------+------+------+--------------------------+
| id | select_type | table    | type  | possible_keys | key              | key_len | ref  | rows | Extra                    |
+----+-------------+----------+-------+---------------+------------------+---------+------+------+--------------------------+
|  1 | SIMPLE      | wp_posts | index | NULL          | type_status_date | 140     | NULL | 5965 | Using where; Using index |
+----+-------------+----------+-------+---------------+------------------+---------+------+------+--------------------------+
1 row in set (0.00 sec)
```

So, this modified query uses same key (_type_status_date_) as original query and query time is practically same.

donnchawp:
@stodorovic - thanks. I added similar code to what you have above and it seems to be working well. Can you give it a spin?

stodorovic:
@donnchawp 

At first look, the code is fine. I'll update websites and we will see how it works in next days.

wp-cache.php
```diff
@@ -3688,16 +3686,29 @@ function wpsc_enable_preload() {
...
+function wpsc_get_post_types() {
+
+	$preload_type_args = apply_filters( 'wpsc_preload_type_args', array(
+		'public'             => true,
+		'publicly_queryable' => true
+	) );
+
+	$post_types = (array) apply_filters( 'wpsc_preload_post_types', get_post_types( $preload_type_args, 'names', 'or' ));
+
+	return join( "', '", array_map( 'esc_sql', $post_types ) );
```
stodorovic: It makes sense to this function returns string with all quotes . I mean:
`'post', 'page',...` instead `post', 'page`. So, I propose:
`return "'" . join( "', '", array_map( 'esc_sql', $post_types ) ) . "'";`

stodorovic:
I found some PHP notices in debug.log:

```
[15-Aug-2017 15:47:40 UTC] PHP Notice:  wpdb::prepare was called <strong>incorrectly</strong>. The query argument of wpdb::prepare() must have a placeholder. Please see <a href="https://codex.wordpress.org/Debugging_in_WordPress">Debugging in WordPress</a> for more information. (This message was added in version 3.9.0.) in /home/frlangusr/public_html_fresh/wp-includes/functions.php on line 4152
[15-Aug-2017 12:06:57 UTC] PHP Notice:  wpdb::prepare est appelée de la mauvaise manière. L’argument de requête wpdb::prepare() doit avoir un paramètre fict
if. Veuillez lire <a href="https://codex.wordpress.org/Debugging_in_WordPress">Débugger dans WordPress</a> (en) pour plus d’informations. (Ce message a été 
ajouté à la version 3.9.0.) in /home/reponsesbio/public_html/wp-includes/functions.php on line 4139
```

wp-cache.php
```diff
@@ -3672,6 +3686,30 @@ function wpsc_enable_preload() {
...
+function wpsc_post_count() {
+	global $wpdb;
+	static $count;
+
+	if ( isset( $count ) ) {
+		return $count;
+	}
+
+	$post_type_list = wpsc_get_post_types();
+	$count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM {$wpdb->posts} WHERE post_type IN ('$post_type_list') AND post_status = 'publish'", $wpdb->posts ) );
```
stodorovic: I optimized query (see PHP notice):

`$count = $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM $wpdb->posts WHERE post_type IN ($post_type_list) AND post_status = %s", 'publish' ) );`
I moved `'` into function `wpsc_get_post_types`. Also `prepare` requires at least one `%s` but we can't use `$post_type_list`. So, add `%s` or remove `prepare`?

wp-cache.php
```diff
@@ -3440,10 +3456,8 @@ function wp_cron_preload_cache() {
 	}
 
 	if ( $wp_cache_preload_posts == 'all' || $c < $wp_cache_preload_posts ) {
-		$types = get_post_types( array( 'public' => true, 'publicly_queryable' => true ), 'names', 'or' );
-		$types = array_map( 'esc_sql', $types );
-		$types = "'" . implode( "','", $types ) . "'";
-		$posts = $wpdb->get_col( "SELECT ID FROM {$wpdb->posts} WHERE ( post_type IN ( $types ) ) AND post_status = 'publish' ORDER BY ID DESC LIMIT $c, 100" );
+		$types = wpsc_get_post_types();
```
stodorovic: According with other changes:
```
$post_type_list = wpsc_get_post_types();
$posts = $wpdb->get_col( $wpdb->prepare( "SELECT ID FROM $wpdb->posts WHERE post_type IN ($post_type_list) AND post_status = %s ORDER BY ID DESC LIMIT %d, 100", 'publish', $c ) );
```

wp-cache.php
```diff
@@ -3688,16 +3686,29 @@ function wpsc_enable_preload() {
 	wp_schedule_single_event( time() + 10, 'wp_cache_full_preload_hook' );
 }
 
+function wpsc_get_post_types() {
+
+	$preload_type_args = apply_filters( 'wpsc_preload_type_args', array(
```
stodorovic: I'm thinking about better name for filter. Maybe `wpsc_preload_get_pt_args` or `wpsc_preload_get_post_types_args` (it seems too long for me, I prefer first variant). Only types isn't totally clear which type?

wp-cache.php
donnchawp: Thanks, I'll remove the prepare() instead as it's not needed.

wp-cache.php
donnchawp: Thanks, I'll add that but leave out the "publish" string, as it's a constant string, not a variable passed to the query.

wp-cache.php
donnchawp: There's not much between them, but I'll go with `wpsc_preload_post_types_args` to look similar to the filter below it.

PR #346, opened 2017-08-15
Don't cache unknown page types, fixes problems with sitemaps

### LINKED ISSUE #325: Site map breaks when user not logged in
I am using sitemap by Yoast SEO. Using WP Super cache plugin breaks sitemap when the administrator is not logged in. 
Sitemap works well when the administrator is logged in to the dashboard.

donnchawp:
Fixes #325
Sitemaps are usually generated by a plugin that exits immediately which
causes problems for caching as it doesn't know what type of page it is.
This code refuses to cache pages if the plugin doesn't know exactly what
type of page is being served. A debug log entry is also generated.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -391,7 +391,7 @@ function wp_cache_ob_callback( $buffer ) {
 	if ( empty( $wp_super_cache_query ) ) {
 		$cache_this_page = false;
 		wp_cache_debug( 'wp_cache_ob_callback: wp_super_cache_query is empty. Not caching unknown page type.' );
-	if ( defined( 'DONOTCACHEPAGE' ) ) {
+	} elseif ( defined( 'DONOTCACHEPAGE' ) ) {
 		wp_cache_debug( 'DONOTCACHEPAGE defined. Caching disabled.', 2 );
 		$cache_this_page = false;
 	} elseif ( $wp_cache_no_cache_for_get && false == empty( $_GET ) && false == defined( 'DOING_CRON' ) ) {
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -371,6 +371,8 @@ function wp_super_cache_query_vars() {
 		$wp_super_cache_query[ 'is_author' ] = 1;
 	if ( is_feed() )
 		$wp_super_cache_query[ 'is_feed' ] = 1;
+
+	return $wp_super_cache_query;
 }
 
 function wp_cache_ob_callback( $buffer ) {
@@ -381,6 +383,14 @@ function wp_cache_ob_callback( $buffer ) {
 
 	// All the things that can stop a page being cached
 	$cache_this_page = true;
+
+	if ( empty( $wp_super_cache_query ) ) {
+		$wp_super_cache_query = wp_super_cache_query_vars();
+	}
+
+	if ( empty( $wp_super_cache_query ) ) {
+		$cache_this_page = false;
+		wp_cache_debug( 'wp_cache_ob_callback: wp_super_cache_query is empty. Not caching unknown page type.' );
 	if ( defined( 'DONOTCACHEPAGE' ) ) {
 		wp_cache_debug( 'DONOTCACHEPAGE defined. Caching disabled.', 2 );
 		$cache_this_page = false;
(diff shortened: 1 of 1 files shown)


stodorovic:
@donnchawp 
I just tried this code:
```php
                elseif ( get_query_var( 'sitemap' ) || get_query_var( 'xsl' ) || get_query_var( 'xml_sitemap' ) ) {
                        $value = "text/xml";
                }
                else { // not a feed
```
before [this line](https://github.com/Automattic/wp-super-cache/blob/8e1f2e65dde94e2f7fba05c9709e5784698f51cd/wp-cache-phase2.php#L1147)

It seems that works. I'll monitor it together with preload testing.

Also, I need to add in .htaccess (if I use rewrite rules) to force PHP:
```
RewriteRule ^sitemap_index.xml$ /index.php [L]
RewriteRule ^([^/]+?)-sitemap([0-9]+)?.xml$ /index.php [L]
RewriteRule ^([a-z]+)?-?sitemap.xsl$ /index.php [L]
```

donnchawp:
Thanks @stodorovic. I may try adding those get_query_var() checks to [wp_super_cache_query_vars()](https://github.com/Automattic/wp-super-cache/blob/8e1f2e65dde94e2f7fba05c9709e5784698f51cd/wp-cache-phase2.php#L352) too and set the is_feed status to true.

stodorovic:
Just notice regarding Yoast SEO. As you already noticed, `template_redirect` isn't triggered. See these lines:

https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-sitemaps.php#L71
https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-sitemaps.php#L235-L236
https://github.com/Yoast/wordpress-seo/blob/trunk/inc/sitemaps/class-sitemaps.php#L191-L194

So, I think that the code in  wp_super_cache_query_vars()  can't be executed.

donnchawp:
Gotcha. This PR calls `wp_super_cache_query_vars()` again if the status array is empty, so along with those query_var checks it should be able to set the is_feed status correctly. I'm testing that.

stodorovic:
Second issue:

I wrote about rewrite rules. So, if I've enabled **Cache Delivery Method** = Expert, wp-super-cache stores files (feeds and sitemaps) as index.html and index.html.gz. So, you can store them as wp-cache  files (keep index.html for HTML content), rewrite rules can't find index.html and PHP will serve correct headers. I'd like to keep rewrite rules for all possible HTML files. (for maximal performance)

Also, please add `X-Robots-Tag` into `$known_headers`. It's important for feeds/sitemaps. (There aren't other way for robots tag).

donnchawp:
I think I just discovered a bug that was introduced in 1.5.0 - Look at [line 1111](https://github.com/Automattic/wp-super-cache/blob/8e1f2e65dde94e2f7fba05c9709e5784698f51cd/wp-cache-phase2.php#L1111) where it's $headers = ... but it should be `$response` (and the line below it should be the same) as that was used originally pre-1.5.0. For Yoast's SEO plugin it doesn't matter as the content type isn't recorded but it'll be fixed in the next release as part of this PR.

By adding the query_var checks the sitemap appears to work ok for me, as a logged in user and logged out. I'll update the PR shortly.

donnchawp:
The index.html and .gz files you saw were created because is_feed() in wp_cache_get_ob() failed to detect a feed (because the sitemap plugin doesn't set it up properly I guess). I replaced is_feed() with my status array and it worked correctly, creating WP-Cache files instead of the index.html/.gz files and the WP-Cache files were reused on the second reload. :)

donnchawp:
With that updated code, the sitemap is now cached as a WP-Cache file. It'll be served by PHP but there's hardly any difference in speed between PHP and mod_rewrite since so little of WordPress is loaded.

stodorovic:
The code seems good. I quickly checked and it works fine. I'll do more testing in next days.

donnchawp:
I'm going to merge this, and sync with WordPress.org SVN to encourage people on the forums to test this. If anything shows up I'll create a new PR.

stodorovic:
I wanted to propose same. I think that we fixed almost everything. I don't see PHP notices and other issues.

PR #352, opened 2017-08-21
Ignore $super_cache_enabled and return correct cache type

donnchawp:
If $super_cache_enabled was disabled the "cache_type" would fallback to "wpcache" which isn't correct because you can't really disable PHP or mod_rewrite modes.


ockham:
Works a treat, and fixes https://horizonfeedback.wordpress.com/2017/08/17/call-for-testing-calypso-wp-super-cache-extension/#comment-392. Thanks a lot!

PR #355, opened 2017-08-22
Fix list markup in advanced-cache.php message

danieljames:
One of the lists was using an 'ol' and writing out the numbers manually,
so they appeared twice. Fix this by removing the manual numbers. Also
convert the other list to an ordered list. And close an unclosed tag.


donnchawp:
Thanks @danieljames, much appreciated!

PR #363, opened 2017-08-24
Don't delete child pages on preload

donnchawp:
Since child pages are most likely to be created after parents pages the
preload process will create the child page before the parent page.
However, it will delete the child pages when it tries to create the
parent page. This fixes that by only deleting files in the directory if
the preloaded post is a page. ref:
https://wordpress.org/support/topic/decending-caching-order-causes-deletion-of-child-pages/


donnchawp:
Confirmed by users that it fixes the problem.

PR #364, opened 2017-08-24
Make sure wp_query is defined before using it.

donnchawp:
Some sites die() before creating $wp_query. This fix creates that object.
It may avoid fatal errors for those users but it's likely the page
type detection will not work.
ref: https://wordpress.org/support/topic/fatal-error-with-latest-update-4/#post-9437870

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -351,7 +351,10 @@ function wp_cache_writers_exit() {
 }
 
 function wp_super_cache_query_vars() {
-	global $wp_super_cache_query;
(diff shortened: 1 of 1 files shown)


donnchawp:
Fixes the problem according to posts on the support forum.

stodorovic:
Hi @donnapep @blowery 

It isn't perfect way. Creating WP Query can open new unpredictable issue. It's much better to check is `$wp_query` is object before calling `wp_super_cache_query_vars`. Please see #367. I've done detail testing with couple plugins and it works without troubles. I'm open for more testing.

PR #368, opened 2017-08-26
Make sure wp_query is defined before using it.

stefb965:
Some sites die() before creating $wp_query. This fix creates that object.
It may avoid fatal errors for those users but it's likely the page
type detection will not work.
ref: https://wordpress.org/support/topic/fatal-error-with-latest-update-4/#post-9437870

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -351,7 +351,10 @@ function wp_cache_writers_exit() {
 }
 
 function wp_super_cache_query_vars() {
-	global $wp_super_cache_query;
+	global $wp_super_cache_query, $wp_query;
+	if ( empty( $wp_query ) ) {
+		$wp_query = new WP_Query();
+	}
 	if ( is_search() )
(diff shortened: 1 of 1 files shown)


stodorovic:
Hi @stefb965 

Please see #364 You created same PR. It's already merged. From other side, it isn't perfect solution. I created #367 which check possible PHP errors and existence of object. (which is faster way)

Only question is it possible to get any useful information when you create object ( also this PR doesn't handle possible occurrence of PHP fatal errors when it's recommended to avoid any complex code).

I spent a lot of time to testing this issue which is consequence of #346. Adding filter at line https://github.com/Automattic/wp-super-cache/blob/master/wp-cache-phase2.php#L395 is better solution. Then user (or custom plugin) can handle these requests (cacheable or no).

Also see Paul's comment:

> * I agree, however, that manually creating the WP_Query object instead of WordPress doing it, isn’t ideal and can lead to potential issues bubbling up elsewhere.

donnchawp:
Already applied, and subsequently reversed in #367.

PR #378, opened 2017-09-06
Add a log viewer to filter unwanted requests from debug log

donnchawp:
Since the debug log will log lots of information we aren't interested in this simple log viewer allows you to filter by various strings (wp-admin, wp-content, wp-json) , and a custom string the user can use as an include or exclude filter.


donnchawp:
It's a good idea, except I wanted the viewer to work even if WordPress had been mangled by something the plugin did and the settings page wasn't working.

Issue #372, opened 2017-08-31
Prevent wp-config changes


mathieuhays:
Would you consider adding a constant or filter that enables us to prevent the plugin from editing the wp-config file?

These changes happens in both `wp_cache_disable_plugin` and `wp_cache_create_advanced_cache` functions in `wp-cache.php`

Thanks :)

donnchawp:
Fixed in #392.

PR #392, opened 2017-10-01
add wpsc_enable_wp_config_edit filter

mathieuhays:
Implements a solution for #372

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -2491,8 +2491,12 @@ function wp_cache_create_advanced_cache() {
 
 	$line = 'define( \'WPCACHEHOME\', \'' . dirname( __FILE__ ) . '/\' );';
 	$should_update_wp_config = apply_filters( 'wpsc_enable_wp_config_edit', true );
+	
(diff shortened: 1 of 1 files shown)


donnchawp:
I made a small change to your patch that added a debug log to the deactivate function. Unfortunately the wp_cache_debug() function isn't available when the plugin is activated so I had to remove it from that after I added it there too..

Can you make one more change to the wp_cache_create_advanced_cache() function? Don't put the check for `$should_update_wp_config` in the file permissions if() - use a new one and print an error message appropriate for it, and return false.

mathieuhays:
Sure, will do

mathieuhays:
I'm not sure what's your process to generate the .po files. 

In case you want the fr_FR translation for the new error message:
`Not allowed to edit %s per configuration.`
`Changement de %s interdit par la configuration.`

donnchawp:
Thanks @mathieuhays - I've merged it as it's so small. 

Translations can be done through the [Translating WordPress site](https://translate.wordpress.org/projects/wp-plugins/wp-super-cache). :)

PR #402, opened 2017-10-06
Make writing to the config file more robust.

donnchawp:
If multiple writers are interacting with the config file at one time
they may corrupt the file so write to a tmp file first and rename.
Also attempt to load the file until successful.

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -2209,37 +2209,56 @@ function wp_cache_replace_line($old, $new, $my_file) {
 	}
 
 	$found = false;
-	$lines = file($my_file);
+	$loaded = false;
+	$c = 0;
+	$lines = array();
+	while( ! $loaded ) {
+		$lines = file( $my_file );
+		if ( ! empty( $lines ) && is_array( $lines ) ) {
+			$loaded = true;
+		} else {
+			$c++;
+			if ( $c > 100 ) {
+				trigger_error( "wp_cache_replace_line: Error  - file $my_file could not be loaded." );
+				return false;
+			}
+		}
+	}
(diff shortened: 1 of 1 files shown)


wp-cache.php
```diff
@@ -2209,37 +2209,56 @@ function wp_cache_replace_line($old, $new, $my_file) {
...
 	foreach( (array)$lines as $line ) {
 		if ( preg_match("/$old/", $line)) {
 			$found = true;
 			break;
 		}
 	}
-	if ($found) {
-		$fd = fopen($my_file, 'w');
+
+	$tmp_file = dirname( $my_file ) . '/' . mt_rand() . '.php';
```
alexsanford: This isn't a big deal, but I wonder if it makes more sense to leave off the `.php` extension here? It won't make any difference, technically, but this function could theoretically be used for more than just php files. What do you think?

If the extension isn't needed, we could even just use `tempnam` instead of creating the filename ourselves. Not a blocker, but it might be a little tidier?

donnchawp: I add the extension to be paranoid about someone requesting the file while it's generating and seen a plain text file with sensitive information. :)

alexsanford: Ah, yes I see.

Another option might be to create the file in `sys_get_temp_dir` instead of `dirname( $my_file )`. Then it shouldn't be accessible to the outside world at all.

I guess I was thinking that `tempnam` did that as well, giving you a full path to a temporary file you could create. But it seems to just give a filename. So probably using `mt_rand` is just as effective for creating a filename, but it might be safer to create the file outside of this publicly accessible directory.

But again, not a blocker for me 😄

Issue #398, opened 2017-10-05
White / Blank Screen


johnw6761:
I am having the white screen issue after turning caching on. I have tried to follow steps others recommend of uninstalling the plugin and reinstalling. 

The issue is there is the uninstall instructions mention editing wp-config and .htaccess however when I open both these files there is no mention of wp super cache so there is nothing to delete?

I remove every trace of the plugin I can find but when I re install I get the same issues.

gui1porpis:
The problem is in **/wp-content/plugins/wp-super-cache/plugins/searchengine.php**, rename and wait for a new fix.

johnw6761:
I tried renaming searchengine.php which didnt work.

I updated the plugin and still didnt work?

donnchawp:
What do you see in your php error log? Do you see any fatal errors?

johnw6761:
No nothing

johnw6761:
Actually this

johnw6761:
Fatal error: Call to undefined function apply_filters() in /websites/123reg/LinuxPackage23/de/16/2_/de162.co.uk/public_html/helenkeeble.co.uk/wordpress/wp-content/plugins/wp-super-cache/wp-cache-phase1.php on line 199

donnchawp:
Oh, what version of WordPress are you using? This might be because you're using an old version of WordPress that doesn't load as much as the current version does. :(

donnchawp:
For the moment, change line 199 to this and it'll work again:
`$rss_types = array( 'application/rss+xml', 'application/rdf+xml', 'application/atom+xml' );`

We'll get another update out soon.

johnw6761:
That worked perfectly thank you!

PR #403, opened 2017-10-06
Update feeds when site updates

donnchawp:
Store a record of archive feed pages that can be deleted when the cache
is updated. This will help keep feeds fresh. Do not record feeds for
single pages. A maximum of 50 feeds is recorded to avoid bloating the
options table.
Will avoid writing to the config file and causing race issues reading
and writing.


donnchawp:
This will also fix #398. The first reply presumed the wrong bug which I fixed in 1.5.7 but it was actually using apply_filters in an old version of WordPress before 4.6. That didn't load apply_filters until much later in the process.

PR #405, opened 2017-10-09
Check for an empty buffer early to avoid processing

donnchawp:
This needs a lot of testing and it could have consequences.


kraftbj:
This makes sense, but understood re additional testing. Going to close for now to allow the team to opt to pick it up later if so desired.

PR #412, opened 2017-10-11
Use $_SERVER[ 'SERVER_PROTOCOL' ] to determine correct HTTP protocol …

robjarsen:
…version for response headers

See: http://php.net/manual/en/function.header.php#92305


donnchawp:
Thanks for the patch!

PR #413, opened 2017-10-11
Use $_SERVER[ 'SERVER_PROTOCOL' ] to determine correct HTTP protocol …

robjarsen:
…version for response headers

See: http://php.net/manual/en/function.header.php#92305

(Last change I needed to propose.)


robjarsen:
@donnchawp

alexsanford:
This looks good to me. It's worth noting that nginx (and probably other webservers as well) seems to automatically change the protocol to HTTP/1.1, regardless of what PHP specifies. But I tested this manually with `php -S` and it works great!

Thanks, @robjarsen !

PR #436, opened 2017-10-27
Hide debug output for REST or Ajax requests

### LINKED ISSUE #79: Debug should not break XML/JSON
We should probably not output any debug code if `DOING_AJAX`, `XMLRPC_REQUEST`, or `REST_REQUEST`.

### LINKED ISSUE #961: Debug should not break XML/JSON
We should probably not output any debug messages if `DOING_AJAX`, `XMLRPC_REQUEST`, or `REST_REQUEST`.

jom:
Fixes Automattic/wp-super-cache#961 

Prevents debug output on XHR or REST API requests.


stodorovic:
I just looked this PR. Function `wp_super_cache_query_vars` is called before `wp_cache_add_to_buffer`. So it's safe to we use `isset( $wp_super_cache_query[ 'is_rest' ] )`.

I think that's better to avoid repeating same code snippet and alternative could be:

```php
if ( strpos( $buffer, '<html' ) === false ||
     isset( $wp_super_cache_query[ 'is_rest' ] ) ||
     isset( $wp_super_cache_query[ 'is_feed' ] )
) {
```

Maybe we can keep comments for is_feed, I'm not sure. Also first condition is true in almost all cases (for XML files).

Related to AJAX request, we don't need anything special because output buffer isn't created. Also, I've created PR Automattic/wp-super-cache#384 and function `wpsc_is_backend` which covers more cases - cron requests, xlmrpc, wp-login, wp-cli and various PHP scripts.

PR #437, opened 2017-10-31
Check that needle is not empty before searching for it.

donnchawp:
If the CDN setting "Exclude if substring" is blank a warning will be
generated when you look for it. This fixes that and uses stripos instead
of stristr as it's faster.
ref: https://wordpress.org/support/topic/another-empty-needle-issue/

### DIFF
--- a/ossdl-cdn.php
+++ b/ossdl-cdn.php
@@ -34,8 +34,8 @@ if ($arr_of_cnames[0] == '') $arr_of_cnames = array();
  * @return Boolean true if to exclude given match from rewriting
  */
 function scossdl_off_exclude_match($match, $excludes) {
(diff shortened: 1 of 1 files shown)


stodorovic:
Just an improvement,, it's little faster (and easier) to use array_filter together with array_map. For example:
`$arr_of_excludes = array_filter( array_map( 'trim', explode( ',', $ossdl_off_exclude ) ) ) ;`

I'm thinking to move this code only on update settings. We should save _array_ instead _string_, then it's possible to we use quicker way for loading - `$ossdl_off_exclude = get_option('ossdl_off_exclude')`. Main trouble is function which converts current options into new format. So, it makes sense if we convert more options on this way.

PR #443, opened 2017-11-06
 Move wp_cache_setting to phase1 file, and load everything on activation

donnchawp:
The function wp_cache_setting() is used by the debug logger in
wp-cache-phase1.php so we need to define that function earlier.
The activation script then has to include the phase1 script so the
plugin can record settings.

Move is_writeable_ACLSafe to phase1 too as it's used there.


donnchawp:
Issue reported here: https://wordpress.org/support/topic/call-to-undefined-function-wp_cache_setting/

PR #449, opened 2017-11-07
Fix/tidy up static page cache admin

donnchawp:


donnchawp:
Weird. The changes to wp-cache-phase1.php belong in #444.

PR #452, opened 2017-11-09
readme overwork

timse201:
some restructure (old text but some deletions in commit 5282efa ) and some styling that works at wordpress and github


stodorovic:
@timse201 Second issue, WP core 4.9.0 has bug which is fixed in 4.9.1. So, plugin  (advanced caching) can't be installed on WP 4.9 because .htaccess can't be modified. Related to #461

timse201:
we can make an upgrade notice
think that is a good way
wp.org does automaticaly update that 4.9 tag to 4.9.1 or 4.9.2 afterwards

stodorovic:
@timse201 Yes, it isn't problem now. It was issue when you have written 4.9.0 (when 4.9.1 isn't yet released), but you didn't read all issues to verify that's correct. I changed it to 4.9.1 in #473

timse201:
issue was 16 days ago

and my commit was 24 days ago

PR #465, opened 2017-11-28
Move functions in phase1 file into phase2 file and load that everywhere

donnchawp:
This will simplify loading the plugin as we don't have to worry about
functions existing before we use them.


donnchawp:
Instead of creating a new file I just used wp-cache-phse2.php because there were only ever functions in that file already. It's then loaded in phase1 and wp-cache.php

stodorovic:
I closed #462 because it's same. I already done some optimizations and I can do it when you merge this PR. Maybe it's better to you merge first this PR and then I'll fix all conflicts in other my PR or close them (and create new).
I quickly tested and it seems OK from my view, but I'll do more test when you merge it.

PR #469, opened 2017-12-05
Fix PHP coding standards

robjarsen:
Part 1

### DIFF
--- a/advanced-cache.php
+++ b/advanced-cache.php
@@ -1,28 +1,31 @@
 <?php
-# WP SUPER CACHE 1.2
-function wpcache_broken_message() {
-	global $wp_cache_config_file;
-	if ( isset( $wp_cache_config_file ) == false )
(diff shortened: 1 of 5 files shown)


stodorovic:
Hi @robjarsen 

@donnchawp and I have replaced all spaces with tabs, but I see that you use _four spaces instead one tab_. There are many occurrences (probably entire PR) and maybe it's better to you create more PRs (you can create more branches in your repository instead of pushing everything into master). For example, each commit could be in separate branch - PR.

donnchawp:
@robjarsen - as @stodorovic  said, any code has to use tabs, not spaces, to indent. Can you fix that? Also, smaller PRs are easier to check and merge than big ones. 

It might turn out that changing the spaces to tabs in this PR will bring the number of modified lines down to make this PR small enough to merge.

Issue #459, opened 2017-11-19
Broken HTML code and other issues (WPSC settings) after merging PR #446


stodorovic:
There are some mistakes. Couple examples for now:

`esc_html_e( 'You or an administrator must disable this. See the <a href="http://php.net/features.safe-mode">safe mode manual page</a> for further details. This cannot be disabled in a .htaccess file unfortunately. It must be done in the php.ini config file.', 'wp-super-cache' );`
Result:
`You or an administrator must disable this. See the &lt;a href=&quot;http://php.net/features.safe-mode&quot;&gt;safe mode manual page&lt;/a&gt; for further details. This cannot be disabled in a .htaccess file unfortunately. It must be done in the php.ini config file.`

`esc_html__( 'A custom url or permalink structure is required for this plugin to work correctly. Please go to the <a href="options-permalink.php">Permalinks Options Page</a> to configure your permalinks.', 'wp-super-cache' );`
Result:
`A custom url or permalink structure is required for this plugin to work correctly. Please go to the &lt;a href=&quot;options-permalink.php&quot;&gt;Permalinks Options Page&lt;/a&gt; to configure your permalinks.`

So, PR #446 replaces `__` with `esc_html__` in many cases when it shouldn't be replaced. Also, it removes `echo` in couple cases and prevent output of some HTML tags.

Screenshot (broken HTML tags):
![wpsc-settings-broken-html1](https://user-images.githubusercontent.com/15647032/33016054-4d85dd28-cded-11e7-8eb3-ac803082528f.PNG)

stodorovic:
Possible solutions could be:

`printf( esc_html__( 'You or an administrator must disable this. See the %1$ssafe mode manual page%2$s for further details. This cannot be disabled in a .htaccess file unfortunately. It must be done in the php.ini config file.', 'wp-super-cache' ) , '<a href="http://php.net/features.safe-mode">', '</a>'));`

or something like this, but it requires changing all translations.

stodorovic:
@timse201 Please help us to fix it. It's consequence of your PR #446

timse201:
yes, will do it today.

i prefer to handle urls outside of translation strings. its safer

stodorovic:
Yes, I agree. And you submit previous PR without detail testing. There are two cases:

- HTML tag is on the end of line. We could add echo or something like this.
[...]

PR #471, opened 2017-12-06
Rename this back as it's needed for WordPress.org

donnchawp:
File was renamed in #452 but plugins on WordPress.org need a readme.txt


timse201:
.txt isnt needed at wp.org. There are many Plugins using md files. Its not really nice but it has styling compatibility for both

donnchawp:
@timse201 are you sure? I looked at the plugins I have installed and a number of them have a readme.md but they also have a readme.txt

stodorovic:
@timse201 readme.txt is needed because it contains: `Tested up to:`, `Stable tag:`, `Requires at least:`. ... Please look https://wordpress.org/plugins/wp-super-cache/ and you will see them on right sidebar.

More info on https://codex.wordpress.org/Writing_a_Plugin:
> # Readme File
> If you want to host your Plugin on https://wordpress.org/plugins/, you need to create a readme.txt file within your plugin directory in a standardized format. You can use the automatic plugin 'readme.txt' generator.

timse201:
https://plugins.trac.wordpress.org/browser/cachify/#tags/2.2.4

stodorovic:
If we are going into this direction, you can see https://github.com/Yoast/wordpress-seo There are both readme.txt and readme.MD

I'm not sure that you way is good. I'd like to we follow codex.wordpress.org. Maybe, I'm wrong, but can you find some documentation about it?

Also, @timse201 Please look #459 Did you test all your PRs? I saw few errors in your code.

timse201:
cant find official documentation but ! i found some tickets

https://meta.trac.wordpress.org/ticket/403

stodorovic:
There are a couple contradictions. First [ticket 403](https://meta.trac.wordpress.org/ticket/403) is reported by Joost de Valk (the founder of Yoast), but Yoast SEO has both (separate) files on github and only readme.txt on wordpress.org (which uses old format).

Also, [ticket 403](https://meta.trac.wordpress.org/ticket/403) recommends `readme.txt`:

> There are no plans to support GitHub flavored markdown. For this reason, we continue to recommend using the name "readme.txt" and not "readme.md", as they are unlikely to be of the same format. Additionally, GitHub uses the readme.md for entirely different purposes than the plugin directory does.

timse201:
sure because there were some issues left.
but these are fixed as i know.
you can upload md file with the new version and check it and if it is not working you can rename it in seconds ;)

PR #484, opened 2017-12-08
PHPCS fixes for wp-cache-config-sample.php

robjarsen:
WPCS fixes via PHPCS.


stodorovic:
It's sample file for _wp-cache-config.php_. Function `wp_cache_replace_line` parses line by line and it isn't possible to use few lines (eg. `Array( ...`).
So, I think that's better to we ignore this file now. Maybe we can improve it in the future, but it could make side effects with current code.

stodorovic:
I'm not sure because `wp_cache_setting` will override these changes. I'm just looking and I think that we need to improve `wp_cache_setting`, `wp_cache_replace_line`, `wp_cache_update_rejected_pages` and probably something more.
I'd like to improve regular expression https://github.com/Automattic/wp-super-cache/blob/master/wp-cache-phase2.php#L1086 before we continue with this file.

stodorovic:
I just created first version - PR #487. It works by my quick testing. Please review it and  test it. I see more places which should be improved, current code isn't perfect, but good for start. I'll continue tomorrow if I find free time.

PR #494, opened 2017-12-13
PHPCS fixes for wp-cache-phase2.php

robjarsen:
WPCS fixes via PHPCS.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -1,22 +1,26 @@
 <?php
 
-function gzip_accepted(){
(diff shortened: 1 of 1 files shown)


stodorovic:
Thanks for PR.

It seems OK, but it's hard to check because there are many lines. Please create few PRs. `wp-cache-phase2.php` has a lot of lines and it's impossible to check entire file in one step. I just created one example which could be easier to check and merge - #495. We could wait merging #495, then submit next PR,... I think that's better approach. Also, it's easier to fix conflicts with other PRs. What you think?

kraftbj:
Going to close this to defer to #754 since the OP has closed their GH account.

PR #516, opened 2018-01-17
Correct the renaming of tmp_wpcache_filename 

KY4U:
Removed unnecessary slash in path which caused renaming to of temp file to fail.  This corrects blank (white) page issues with cached pages with GET parameters.


donnchawp:
Thanks for spotting this. What operating system or path caused an error when renaming with an extra slash?

KY4U:
This was on Microsoft Windows (Azure webapp).  I cannot provide a specific path at this time as the log file has been overwritten.

PR #519, opened 2018-01-31
Post Update now clears category cache

clementduncan:
If a post was updated, it would clear the homepage and post cache, but the category pages wouldn't clear.
> category/stories/...

The function was set to only update categories when a post was changed from un-published to published or vice versa.
It's been changed so it clears the category cache when a post goes from un-published to published or a change is made when it's published.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -2587,7 +2587,7 @@ function wpsc_post_transition( $new_status, $old_status, $post ) {
 	if (
 		($old_status == 'publish' && $new_status != 'publish' ) // post unpublished
 		||
-		($old_status != 'publish' && $new_status == 'publish') // post published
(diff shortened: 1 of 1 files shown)


stodorovic:
I've added other way in #577. Function `wp_cache_post_edit` calls `wpsc_delete_post_archives`. To avoid double processing of same post, I reverted these changes in #595 and I'm calling  `wp_cache_post_edit` if post is just published . I'm testing it and it seems fine.

I've found new issue when taxonomy is removed from a post - cache for this taxonomy isn't cleared. I'll try to create new PR which fixes it. It's possible to use hook [`set_object_terms`](https://developer.wordpress.org/reference/hooks/set_object_terms/).

@clementduncan It would be great if you could test #595 and see is it working correctly.

PR #537, opened 2018-04-02
adding hook for full cache purges

futtta:
Goal: allow other plugins to hook into WP Super Cache's clearing the cache entirely by adding a do_action

Reason: in my case I want to purge Autoptimize's cache when WP SC's is fully cleared. 

cfr. https://github.com/Automattic/wp-super-cache/issues/903

Will be happy to improve when/ if needed.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -2094,10 +2094,10 @@ function wp_cache_phase2_clean_cache($file_prefix) {
 						@unlink( $blog_cache_dir . 'meta/' . $file );
 					}
 				}
-				do_action('wp_cache_cleared');
 			}
 		}
 		closedir($handle);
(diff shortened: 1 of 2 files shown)


donnchawp:
It's a good idea, but the change at line 2097 of wp-cache-phase2.php will run the action in a while loop. It should go a few lines down by the wp_cache_writers_exit().

futtta:
Wouldn't it be better to do so a bit higher up, inside the if-block just after the while loop gets closed to ensure the action only fires if indeed the cache was purged? else it _could_ happen that the action is called even if the cache was not purged due to `$handle = @opendir( $blog_cache_dir )` not being true?

futtta:
so should we also add the same do_action elsewhere for when the supercache
directories get cleared?

PR #549, opened 2018-04-30
Fix plugin name - WP Crontrol

garrett-eclipse:
In alert and readme was WP Control so I corrected to WP Crontrol


donnchawp:
Thanks Garrett, never even noticed that!

garrett-eclipse:
No worries, definitely easy to overlook. I only spotted it cause I use it and always mistype it myself

PR #554, opened 2018-05-15
Update version to 1.6.0 and changelog.

donnchawp:


donnchawp:
'course that should be 1.6.1..

Issue #522, opened 2018-02-12
Why vary on cookie?


ahmed-sigmalux:
WP Super Cache is hardcoded to add this header:

`header( "Vary: Accept-Encoding, Cookie" );`

In general, varying on cookies is bad practice, so why does WP Super Cache add vary on cookie?

Currently I'm removing the header and only varying on Accept-Encoding, but I'd like to know if there's some super secret and good reason for WP Super Cache to vary on cookies.

marcotini:
I noticed that too. It's wrong because it's create double option on Vary and nginx create one too.

donnchawp:
The plugin can cache pages for logged in users and those users are identified by cookie. If you remove "cookie" from there I think you may see odd things happen in proxied connections as a proxy/caching server might cache a private page.

But if you can test it and provide evidence that it's not needed it can be removed.

donnchawp:
Unfortunately I don't have a host running Nginx to test on. Can you tell Nginx not to send that Vary header? A [comment here](https://github.com/wagtail/wagtail/issues/1111) says how to do it.

Or you could test if the Vary header has already been sent by Nginx using the PHP headers_list() command, but you'll need to edit WP Super Cache code to do that. Line [161 of wp-cache-phase2.php](https://github.com/Automattic/wp-super-cache/blob/71b92a6ce57a7f73971bc8b80ecb1df527ec61ec/wp-cache-phase2.php#L161) is probably where you want to add that test.

Do you know why Nginx adds the Vary: Cookie header? Is it for caching the web server does?

marcotini:
Nginx doesn’t add Cookie, but Accept-Encoding. 
> Can you tell Nginx not to send that Vary header? A comment here says how to do it.
> 

Yes I can, but the Accept-Encoding? Who show it?

donnchawp:
Check out #555 which lets you change that header by defining the constant WPSC_VARY_HEADER in your wp-config.php. It should have the header part after "Vary:".

PR #555, opened 2018-05-15
Allow the Vary and Cache-Control headers to be modified.

### LINKED ISSUE #523: Please add filter / setting to modify cache-control headers
WP Super Cache hardcodes the cache-control headers to:

`header( "Cache-Control: max-age=3, must-revalidate" );`

A max-age of 3 seconds is extremely short and unhelpful for most websites.

Can you add a filter / setting that allows users to modify the max-age value?

donnchawp:
Fixes #523 and #522
Define the constants WPSC_VARY_HEADER and WPSC_CACHE_CONTROL_HEADER to
override the default Vary and Cache-Control headers.
If the Vary header doesn't include Accept-Encoding visitors may get
mixed compressed or not pages.
If it is missing Cookie the wrong file might be sent to anonymous users.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -158,8 +158,20 @@ function wp_cache_serve_cache_file() {
 			if ( isset( $wp_cache_disable_utf8 ) == false || $wp_cache_disable_utf8 == 0 )
 				header( "Content-type: text/html; charset=UTF-8" );
 
-			header( "Vary: Accept-Encoding, Cookie" );
-			header( "Cache-Control: max-age=3, must-revalidate" );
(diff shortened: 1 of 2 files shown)


kasperkamperman:
Any documentation on this point. How do I add this to my wp-config?

WPSC_VARY_HEADER = true? Or something?

donnchawp:
You define WPSC_VARY_HEADER and WPSC_CACHE_CONTROL_HEADER by setting the string that should be sent for each header. For example:
`define( 'WPSC_VARY_HEADER', 'Accept-Encoding, Cookie' );`
`define( 'WPSC_CACHE_CONTROL_HEADER', 'max-age=3, must-revalidate' );`

If you set them to a blank string then that header won't be used.

PR #557, opened 2018-05-17
Fix #556

### LINKED ISSUE #556: Erroneous "Rewrite rules must be updated" notice that never goes away
It keeps saying:

The rewrite rules required by this plugin have changed or are missing. Scroll down the Advanced Settings page and click the **Update Mod_Rewrite Rules** button.

Clicking the button doesn't help (even though rewrite rules are updated/changed).

This is caused by https://github.com/Automattic/wp-super-cache/commit/6943fe9ea8fcb027aba37301d0ac65f99ba14ade#diff-d5e23b1d94909d6b8f2281f18d119027R377 -- that condition looks like it should be inverted.

zytzagoo:
Inverting the condition in order to actually use a proper path if it's not already set. Fixes #556

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -378,7 +378,7 @@ function wp_cache_manager_error_checks() {
 	}
 
 	if ( ! $is_nginx && function_exists( "is_main_site" ) && true == is_main_site() ) {
-	if ( isset( $htaccess_path ) ) {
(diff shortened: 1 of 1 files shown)


donnchawp:
Sets the path to the .htaccess file correctly on a multisite when on the Advanced settings page on the main site.

PR #561, opened 2018-06-09
Fix install for enviroment with open_basedir

unreturned:
Hello, please, use https://developer.wordpress.org/reference/functions/get_temp_dir/ for temporary directory, not clear sys_get_temp_dir, because sys_get_temp_dir return mostly /tmp and ignore users tmp directory and setting open_basedir

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -1055,7 +1055,7 @@ function wp_cache_replace_line( $old, $new, $my_file ) {
 		return false;
 	}
 
-	$tmp_file = tempnam( sys_get_temp_dir(), 'wpsc' );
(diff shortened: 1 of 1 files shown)


unreturned:
Problem was intrduced by https://github.com/Automattic/wp-super-cache/commit/dec9ffce1d10e8868c18e600b3338d448d931268

unreturned:
More customers affected by this problem https://wordpress.org/support/topic/error-in-v1-6-1-in-shared-hosting-environments-open_basedir-restriction/

unreturned:
@donnchawp Please, review it. Thanks in advance.

donnchawp:
I fixed this in #563 by writing (again) directly to the config file. Doing that doesn't need write access to the wp-content directory so avoiding this problem.

PR #562, opened 2018-06-11
Fix/avoid serving expired file

viniciusbarros:
I noticed this issue when the cache is served directly from the file. So all I added was a check to compare the file creation time + $cache_max_time against current time.

If expired, we return false and new cache is generated.

Notes:
- I came across this issue when storing cache files in a custom directory. E.g: `/tmp/wp-super-cache`
- I tested it locally and it worked as expected.


donnchawp:
Thanks. Good catch. I fixed some problems with space characters in your code.

PR #566, opened 2018-06-16
Fixes typo in error message.

emrikol:
I believe this is a typo, from what I can tell.  If not, please disregard :)


donnchawp:
Thanks! :)

PR #574, opened 2018-07-24
Make it easier to add new WP Super Cache plugins by outside code.

donnchawp:
Adding new plugins to WP Super Cache to extend it is complicated but
where they should go should be easy. This patch makes the plugin check
in a particular directory if it exists and load the PHP files there.
The files there won't be deleted when WP Super Cache is updated.


donnchawp:
wpsc_add_plugin() and wpsc_delete_plugin() can be used to add new plugins to be loaded by WP Super Cache.
Each takes one parameter, a filename. It can be with or without the ABSPATH but ABSPATH is stripped out before storing the filename.
The files are loaded way before most of WordPress loads.

joloco:
I know this is an old commit, but this is exactly the thing I was looking for!

However, something isn't quite right, and I wonder if the desired behaviour is that of the comment, or the code.

The comment states that the code will _"also look for plugins in `wp-content/wp-super-cache-plugins/`"_ but this is incorrect.

The code instead looks for plugins in `wp-content/plugins/wp-super-cache-plugins/` which is, of course, inside the general WP `plugins` directory.

This does work, but it seems a bit strange, as I'd have thought that the WP `plugins` directory was intended only for proper WP plugins.

There's also the (small) risk that someone could add a new plugin to the WP plugins database, called `wp-super-cache-plugins`, which when installed would over-write any plugins added by this commit.

PR #575, opened 2018-07-25
Add warning about DISABLE_WP_CRON

### LINKED ISSUE #532: HTTP-Request to wp_cron.php failed when DISABLE_WP_CRON true
When DISABLE_WP_CRON  is true than http request for wp-cron failes (because the response is empty when disabled). The request is initiated by https://github.com/Automattic/wp-super-cache/blob/master/wp-cache.php#L263

image

WP Super Cache 1.5.9
Wordpress 4.9.4

donnchawp:
Fixes #532

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -242,8 +242,14 @@ function wp_cache_manager_error_checks() {
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


danielkesselberg:
Thank you

PR #580, opened 2018-08-07
Add cookie helper functions to modify the cache key more easily.

donnchawp:
Inspired by https://wordpress.org/support/topic/php-cookies-doesnt-work-with-wp-super-cache/
This will make it much easier for plugin authors that use cookies to
identify visitors to integrate with WP Super Cache.
Use the function wpsc_add_cookie( $name ) to add a cookie name to the
list of cookies and wpsc_delete_cookie( $name ) to remove it.

Example code to add the cookie "euCookie" used by the EU Cookie Law
plugin;
https://wordpress.org/plugins/eu-cookie-law/

--------------
function add_wpsc_cookies() {
    if ( function_exists( 'wpsc_add_cookie' ) ) {
        wpsc_add_cookie( 'euCookie' );
    }
}

add_action( 'init', 'add_wpsc_cookies' );
--------------

It's wrapped in an action on "init" to make sure that WP Super Cache is
loaded but the wpsc_add_cookie() function only needs to be called once.
Calling it multiple times does not add multiple entries for the same
cookie name.


andreasciamanna:
I see an action more appropriate for adding a cookie key.

In ky opinion, it's also cleaner (no need to check if the function exists).

I'll be happy to implement it in the next WPML bugfix version, by the way.

donnchawp:
I thought about implementing it as a filter first but since the cookie key is only supposed to be added once I wanted to get away from having code that runs unnecessarily each time a plugin runs.

However, it is more convenient. I'll create a new PR with such a filter (and while I'm at it do the same for loading plugins)

donnchawp:
#582 will add the filter you need. Does it look ok? Is "init" too early to fire that filter?

PR #582, opened 2018-08-09
Add wpsc_cookies and wpsc_plugins filters to modify those settings

donnchawp:
These filters will allow WordPress plugins to modify the cookies and
plugins lists used by WP Super Cache instead of the add/delete functions
introduced in #574 and #580.
Duplicate entries are removed so the filter function does not need to
worry about that. The filters fire on 'init'.

Example filter code:

```
function myplugin_add_wpsc_cookies( $cookies ) {
    $cookies[] = 'myplugin';
    return $cookies;
}
add_filter( 'wpsc_cookies', 'myplugin_add_wpsc_cookies' );
```

```
function myplugin_delete_wpsc_cookies( $cookies ) {
    if ( in_array( 'myplugin', $cookies ) ) {
        unset( $cookies[ array_search( 'myplugin', $cookies ) ] );
    }
    return $cookies;
}
add_filter( 'wpsc_cookies', 'myplugin_delete_wpsc_cookies' );
```

As suggested by @johnbillion and @andreasciamanna in #580.

You can also use `do_action` to add a cookie like this:
```
do_action( 'wpsc_add_cookie', 'euCookie' );
```
Use `wpsc_delete_cookie` to remove the cookie.

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -83,9 +83,24 @@ if ( class_exists( 'WP_REST_Controller' ) ) {
 }
 
 function wp_super_cache_init_action() {
+	global $wpsc_cookies, $wpsc_plugins;
+
 	load_plugin_textdomain( 'wp-super-cache', false, basename( dirname( __FILE__ ) ) . '/languages' );
 
 	wpsc_register_post_hooks();
+
+	$cookies = array_unique( apply_filters( 'wpsc_cookies', $wpsc_cookies ) );
+	if ( $cookies != $wpsc_cookies ) {
+		$wpsc_cookies = $cookies;
+		wp_cache_setting( 'wpsc_cookies', $wpsc_cookies );
+	}
+
+	$plugins = array_unique( apply_filters( 'wpsc_plugins', $wpsc_plugins ) );
+	if ( $plugins != $wpsc_plugins ) {
+		$wpsc_plugins = $plugins;
+		wp_cache_setting( 'wpsc_plugins', $wpsc_plugins );
+	}
+
 }
 add_action( 'init', 'wp_super_cache_init_action' );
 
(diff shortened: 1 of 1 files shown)


andreasciamanna:
As I've tried to explain [here](https://github.com/Automattic/wp-super-cache/pull/580#issuecomment-411411300) I think an action is more appropriate and safer than a filter.

If properly used, a hook to the filter could mess with the content of `$cookies`.

An action seems to me more straight-forward.

In a plugin, I could call `do_action( 'wpsc_add_cookie', 'my-plugin-cookie-name' );`

And in wp-super-cache, `wpsc_add_cookie()` could be hooked to that action:

```php
function wpsc_add_cookie( $name ) {
	global $wpsc_cookies;
	if (
		! isset( $wpsc_cookies ) ||
		! is_array( $wpsc_cookies ) ||
		! in_array( $name, $wpsc_cookies )
	) {
		$wpsc_cookies[] = $name;
		wp_cache_setting( 'wpsc_cookies', $wpsc_cookies );
	}
}
add_action( 'wpsc_cookies' );
```

donnchawp:
Ah. I wondered about that. Your example makes sense and avoids running an apply_filters() on every page load if there's no outside code that wants to modify the cookie list. Thanks. I'll modify the PR.

donnchawp:
I updated the PR, changing the filters to actions so you can do `do_action( 'wpsc_add_cookie', 'euCookie' );' to add a cookie called 'euCookie'.

andreasciamanna:
Thanks a lot!

We are a bit at the end of our development cycle, but I'll see if I can manage to include it in the next WPML bugfix release. 

If not, it will be included in the following one.

andreasciamanna:
Just tested it with `add_cookie_plugin_filters` branch and it works perfectly.

Found a bug in `wp-cache-phase2.php` though: I'll create a PR.

donnchawp:
Great. that's merged and I'll get a new release out this evening or tomorrow.

PR #583, opened 2018-08-09
Fix strncmp missing argument

andreasciamanna:
Before this fix, I was getting many `PHP Warning:  strncmp() expects exactly 3 parameters, 2 given in wp-cache-phase2.php on line 3075` as soon as any cookie was created.

I believe the original intention of using `strncmp` instead of `strcmp` was a question of performance.

As per my knowledge, `strpos` should be faster, especially when comparing the beginning of a string.

In any case, I'm keeping the use of `strncmp` and pass the missing argument which, I believe, should be `5` (i.e., the length of `HTTP_ `).

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -3072,7 +3072,7 @@ if ( ! function_exists( 'apache_request_headers' ) ) {
 		$headers = array();
 
 		foreach ( array_keys( $_SERVER ) as $skey ) {
-			if ( 0 === strncmp( $skey, 'HTTP_' ) ) {
+			if ( 0 === strncmp( $skey, 'HTTP_', 5 ) ) {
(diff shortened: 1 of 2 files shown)


stodorovic:
It's my mistake (I wanted to use strpos). It's fallback for [`function apache_request_headers`](http://php.net/manual/en/function.apache-request-headers.php) which doesn't exist in all installations. I've tested it and I don't remember which is better. I agree with you that [strpos](http://php.net/manual/en/function.strpos.php) is more clear solution.

I see other lines which aren't related to strncmp (lines from other commits). Anyway I just created #584 which replaces strncmp with strpos.

andreasciamanna:
Sorry for the other changes: they weren't supposed to be included.

Anyway, as your PR is more in my cords, I'll close this one :)

donnchawp:
Thanks @andreasciamanna - I'll merge #584 :)

PR #612, opened 2018-09-14
Updated Supercache domain mapping plugin so that domain-mapped sites …

sjeng:
…would have their cache cleared correctly after content update.

The expected behavior is that after updating an already published content, cache should be cleared.  But the behavior was broken by #595.

It appears that in version 1.6.4, clearing existing page cache logic is handled exclusively `wp_cache_post_id_gc`, which uses `get_current_url_supercache_dir` to get the directory path to the cached files.  However, the domain mapping plugin does not hook into `get_current_url_supercache_dir`, so the non-domain-mapped directory path is used, which is not where the cached files lives.

### DIFF
--- a/plugins/domain-mapping.php
+++ b/plugins/domain-mapping.php
@@ -35,24 +35,47 @@ function domain_mapping_gc_cache( $function, $directory ) {
 
 function domain_mapping_supercachedir( $dir ) {
 	global $cache_path;
-	if ( ! function_exists( 'domain_mapping_warning' ) ) {
-		return $dir;
-	}
-
-	$siteurl = domain_mapping_siteurl( false );
+	$siteurl = wp_supercache_domain_mapping_siteurl();
 	if ( ! $siteurl ) {
 		return $dir;
 	}
 
+	return trailingslashit( $cache_path . 'supercache/' . $siteurl );
+}
+
+function domain_mapping_current_url_supercachedir( $dir ) {
+	global $WPSC_HTTP_HOST;
+	$siteurl = wp_supercache_domain_mapping_siteurl();
+	if ( ! $siteurl ) {
+		return $dir;
+	}
+
+	return str_replace(	$WPSC_HTTP_HOST, $siteurl, $dir );
+
+}
+
+function wp_supercache_domain_mapping_siteurl() {
+
+	if ( ! function_exists( 'domain_mapping_warning' ) ) {
+		return false;
+	}
+
+	$siteurl = domain_mapping_siteurl( false );
+	if ( ! $siteurl ) {
+		return false;
+	}
+
 	$protocol = ( isset( $_SERVER['HTTPS'] ) && 'on' === strtolower( $_SERVER['HTTPS'] ) ) ? 'https://' : 'http://';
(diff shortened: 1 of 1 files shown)


stodorovic:
I'm not sure that's good solution because filter `supercache_dir` could be triggered from function `wp_cache_serve_cache_file` when object `$wpdb` doesn't exist ( function `domain_mapping_siteurl` uses WP database). I'll check is it possible to find better workaround. Anyway we should modernize function `wp_cache_post_change`.

stodorovic:
Thank you for reporting this issue, but I can't reproduce it (I've installed latest WP and WP Super Cache plugin - master repository, Centos/Apache ).

You are trying to add filter which will be triggered to [this line](https://github.com/Automattic/wp-super-cache/blob/1.6.4/wp-cache-phase2.php#L2801), but you can see that the function  `wp_cache_post_id_gc` triggers `do_action( 'gc_cache', 'prune', $permalink );`. So, we don't need new filters and it works in my test ( `$permalink` points to correct directory in my case ). I think that we have to fix function `domain_mapping_gc_cache` if you see an issue. Could you enable debug log and check values in function domain_mapping_gc_cache? ( `$permalink` points to correct directory in my case )

It's possible that you have issue if URL contains query string. Could you send more details how to reproduce it? Possible conflicts with third party plugins?

Also, your patch doesn't work correctly for me. Example (from my test):
```
'example.com/site2/2018/09/16/test/' => 'site2.example.com//site2/2018/09/16/test/'
```
As you see it isn't correct directory and `str_replace( $WPSC_HTTP_HOST, $siteurl, $dir );` doesn't seem as good solution.

sjeng:
Thanks for looking into this so quickly! 

Strangely enough I am also unable to recreate this issue locally right now, but it is currently an issue on our production environment.  I am going to try to trace through the code again and see what exactly is happening.  Once I figure out the exact cause, I'll test out the changes you made and see if it addresses the issue.

sjeng:
I think I figured out what's happening here.  

`domain_mapping_gc_cache` is explicitly deleting 2 files when pruning a domain mapped cache directory:  index.html and index.html.gz.  The actual cached filename is index-https.html, so by adding that to the list of files to be deleted fixed the issue.

I think the reason why this issue surfaced is because of the recent change to `wpsc_post_transition`, where before the change, updated published content's cached files were being deleted by `wpsc_delete_url_cache`.  

Since the change, `wpsc_delete_url_cache` isn't called for updating content, so the job of clearing domain mapped cached files fell solely on the domain mapping plugin (rightly so).  But of course since the plugin only deletes specific files, instead of all files in the directory, we're running into issue with https files not being cleared properly.

So TLDR version:  we can simply fix the issue by updating `domain_mapping_gc_cache` so that https version of the files gets cleared out properly.  I'll close this PR since it doesn't actually solve the root of the issue.

stodorovic:
Thank you for feedback! I just updated #615. It isn't perfect but for now it's acceptable. Could you test it?

sjeng:
#615 looks good!  I tested the change and https cached files were correctly cleared after update.  Thanks!

PR #633, opened 2018-11-24
Fix erratic $log_file_link and $raw_log_file_link on Debug TAB on Windows

DrLightman:
On Windows backslashes in path break the replace. For example **$log_file_link** may become something like this:

`http://example.org/D:/my_sites/example_org/wp-content/cache/view_e1b50ae15624d52670a4a76c4ce03610.php`

Because ABSPATH value on Windows is `D:\my_sites\example_org/` therefore it doesn't get replaced with the empty string.

Same goes with **$raw_log_file_link** since they are being generated in the same manner involving a string replacement.


stodorovic:
@DrLightman  Thanks. I think that's better to we keep _str_replace_. Anyway, there are two variants which could work:
* Normalization of `ABSPATH`:
`$abspath = functions_exists( 'wp_normalize_path' ) ? wp_normalize_path( ABSPATH ) : str_replace( '\\', '/', ABSPATH );`

* Get realpath for `$cache_path` before _str_replace_ (it seems as more clear solution):
`$cache_rel_path = str_replace( ABSPATH, '', wpsc_get_realpath( $cache_path ) );`
`$log_file_link = "<a href='" . site_url( "{$cache_rel_path}view_{$wp_cache_debug_log}" ) .`...


I use [cygwin](https://www.cygwin.com) on Windows and I can't test it at this moment. I'd like to we avoid [`substr`](http://php.net/manual/en/function.substr.php)(on this way) if it's possible. Could you please check is it working on Windows and update PR?

DrLightman:
Hi @stodorovic can you check now? I wasn't able to make it shorter, I had to start with:

**ABSPATH** = `D:\Sites\example_org/`
**$cache_path** = `D:\Sites\example_org\wp-content/cache/`

Also, I don't know if this kind of replacement might break when an user customizes the cache location in the Advanced tab... by moving it outside the document root, is it even possibile, theoretically?

wp-cache.php
```diff
@@ -2085,8 +2085,16 @@ function wp_cache_debug_settings() {
...
-	$log_file_link = "<a href='" . site_url( str_replace( ABSPATH, '', "{$cache_path}view_{$wp_cache_debug_log}" ) ) . "'>$wp_cache_debug_log</a>";
-	$raw_log_file_link = "<a href='" . site_url( str_replace( ABSPATH, '', "{$cache_path}{$wp_cache_debug_log}" ) ) . "'>" . __( 'RAW', 'wp-super-cache' ) . "</a>";
+        if ( function_exists( 'wp_normalize_path' ) ) {
+		$cache_url = wp_normalize_path( $cache_path );
+		$cache_url = str_replace( wp_normalize_path( ABSPATH ), '', $cache_url );
+	} else {
+		$cache_url = str_replace( '\\', '/', $cache_path );
+		$cache_url = str_replace( str_replace( '\\', '/', ABSPATH ), '', $cache_url );        
+	}
+	$cache_url = site_url( $cache_url );
```
stodorovic:
```suggestion
	$cache_url = trailingslashit( site_url( $cache_rel_path ) );
```
It's possible to realpath strips last slash.

stodorovic:
I've checked on older PHP versions (windows) and it seems that we could reduce complexity. Could you check my suggestions? Is it working on your installations?

PR #636, opened 2018-12-07
Promote the Jetpack Site Accelerator on the CDN page.

donnchawp:
The free site accelerator is a part of Jetpack which is commonly
installed on WordPress blogs so we should let users know about it.

### DIFF
--- a/ossdl-cdn.php
+++ b/ossdl-cdn.php
@@ -272,13 +272,44 @@ function scossdl_off_options() {
 	$example_cdn_uri .= '/wp-includes/js/jquery/jquery-migrate.js';
 	$example_cdn_uri  = esc_url( $example_cdn_uri );
 	?>
+		<h3><?php _e( 'Simple CDN' ); ?></h3>
 		<p><?php _e( 'Your website probably uses lots of static files. Image, Javascript and CSS files are usually static files that could just as easily be served from another site or CDN. Therefore, this plugin replaces any links in the <code>wp-content</code> and <code>wp-includes</code> directories (except for PHP files) on your site with the URL you provide below. That way you can either copy all the static content to a dedicated host or mirror the files to a CDN by <a href="https://knowledgelayer.softlayer.com/faq/how-does-origin-pull-work" target="_blank">origin pull</a>.', 'wp-super-cache' ); ?></p>
 		<p><?php printf( __( '<strong style="color: red">WARNING:</strong> Test some static urls e.g., %s  to ensure your CDN service is fully working before saving changes.', 'wp-super-cache' ), '<code>' . esc_html( $example_cdn_uri ) . '</code>' ); ?></p>
 
 	<?php if ( get_home_url() !== $ossdl_off_blog_url ) { ?>
 		<p><?php printf( __( '<strong style="color: red">WARNING:</strong> Your siteurl and homeurl are different. The plugin is using %s as the homepage URL of your site but if that is wrong please use the filter "ossdl_off_blog_url" to fix it.', 'wp-super-cache' ), '<code>' . esc_html( $ossdl_off_blog_url ) . '</code>' ); ?></p>
 	<?php } ?>
-
+		<h3><?php _e( 'Jetpack CDN' ); ?></h3>
+		<p><?php printf(
+			__( 'The free %1$sJetpack plugin%2$s has a %3$sSite Accelerator%2$s feature that is better than the CDN in this plugin as it will offload requests for static files off your server. ', 'wp-super-cache' ),
+			'<a href="https://jetpack.com/">',
+			'</a>',
+			'<a href="https://jetpack.com/support/site-accelerator/">'
+		); ?></p>
+	<?php
+	if ( class_exists( 'Jetpack' ) ) {
+		if ( Jetpack::is_module_active( 'photon' ) ) {
+			?><p><?php printf(
+				__( 'You already have Jetpack installed and %1$sSite Accelerator%2$s enabled on this blog. <strong>We recommend you do not use the CDN on this page.</strong>', 'wp-super-cache' ),
+				'<a href="https://jetpack.com/support/site-accelerator/">',
+				'</a>'
+			); ?></p><?php
+		} else {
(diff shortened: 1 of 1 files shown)


ossdl-cdn.php
```diff
@@ -272,13 +272,44 @@ function scossdl_off_options() {
...
 		<p><?php printf( __( '<strong style="color: red">WARNING:</strong> Test some static urls e.g., %s  to ensure your CDN service is fully working before saving changes.', 'wp-super-cache' ), '<code>' . esc_html( $example_cdn_uri ) . '</code>' ); ?></p>
 
 	<?php if ( get_home_url() !== $ossdl_off_blog_url ) { ?>
 		<p><?php printf( __( '<strong style="color: red">WARNING:</strong> Your siteurl and homeurl are different. The plugin is using %s as the homepage URL of your site but if that is wrong please use the filter "ossdl_off_blog_url" to fix it.', 'wp-super-cache' ), '<code>' . esc_html( $ossdl_off_blog_url ) . '</code>' ); ?></p>
 	<?php } ?>
-
+		<h3><?php _e( 'Jetpack CDN' ); ?></h3>
+		<p><?php printf(
+			__( 'The free %1$sJetpack plugin%2$s has a %3$sSite Accelerator%2$s feature that is better than the CDN in this plugin as it will offload requests for static files off your server. ', 'wp-super-cache' ),
```
dbtlr: `that is better than the CDN in this plugin` -> `that is may be preferable to many traditional CDN services offered in this plugin`

ossdl-cdn.php
```diff
@@ -272,13 +272,44 @@ function scossdl_off_options() {
...
+			'<a href="https://jetpack.com/">',
+			'</a>',
+			'<a href="https://jetpack.com/support/site-accelerator/">'
+		); ?></p>
+	<?php
+	if ( class_exists( 'Jetpack' ) ) {
+		if ( Jetpack::is_module_active( 'photon' ) ) {
+			?><p><?php printf(
+				__( 'You already have Jetpack installed and %1$sSite Accelerator%2$s enabled on this blog. <strong>We recommend you do not use the CDN on this page.</strong>', 'wp-super-cache' ),
```
dbtlr: `You already have Jetpack installed and %1$sSite Accelerator%2$s enabled on this blog. <strong>We recommend you do not use the CDN services on this page, as they may be redundant.</strong>`

ossdl-cdn.php
```diff
@@ -272,13 +272,44 @@ function scossdl_off_options() {
...
+			?><p><?php printf(
+				__( 'You already have Jetpack installed but %1$sSite Accelerator%2$s is disabled on this blog. Enable it on the %3$sJetpack settings page%2$s.', 'wp-super-cache' ),
+				'<a href="https://jetpack.com/support/site-accelerator/">',
+				'</a>',
+				'<a href="' . admin_url( 'admin.php?page=jetpack#/settings' ) . '">'
+			); ?></p><?php
+		}
+	} else {
+			?><p><?php printf(
+				__( '%1$sJetpack%2$s is not installed on your site. The Site Accelerator feature is free to use on any WordPress site. You should give it a go!', 'wp-super-cache' ),
```
dbtlr: `%1$sJetpack%2$s is not installed on your site. The Site Accelerator feature is free to use and offers the same benefit as other CDN services, on any WordPress site. You should give it a try!`

ossdl-cdn.php
prhomhyse: I thought of this as WordPress 5.0 shipped but from [JetPack's docs](https://jetpack.com/support/site-accelerator/) 

> No cache invalidations – currently the images are cached “forever” and static assets will be tied >to the public version of WordPress, Jetpack, or WooCommerce that you’re using. For images, if >you want to “refresh” an image you will need to change its file name. Adding random query >arguments, commonly known as cachebusters, will not work.

That's a serious limitation to bear in mind while recommending it to more than a million installs of WPSC. 
Otherwise, this is a great idea

dbtlr: That's a really good point. Maybe we should try to make the callout a little less "you should do it" and more "here's another (potentially easier) option"

donnchawp: I agree. The wording on the CDN page should make reference to this limitation of Site Accelerator. I'll update it.

donnchawp: I updated the language, and also disabled the CDN if SA is active in https://github.com/Automattic/wp-super-cache/pull/636/commits/0b641bf8407ad78adcc7a0a46e98ab6d529817e9

ossdl-cdn.php
```diff
@@ -272,13 +276,47 @@ function scossdl_off_options() {
...
+			?><p><?php printf(
+				__( 'You already have Jetpack installed but %1$sSite Accelerator%2$s is disabled on this blog. Enable it on the %3$sJetpack settings page%2$s.', 'wp-super-cache' ),
+				'<a href="https://jetpack.com/support/site-accelerator/">',
+				'</a>',
+				'<a href="' . admin_url( 'admin.php?page=jetpack#/settings' ) . '">'
+			); ?></p><?php
+		}
+	} else {
+			?><p><?php printf(
+				__( '%1$sJetpack%2$s is not installed on your site. The Site Accelerator feature is free to use on any WordPress site and offers the same benefit as other CDN services. You should give it a try!', 'wp-super-cache' ),
```
beaulebens: This one might be better as something like "Jetpack's Site Accelerator feature is free to use...".

Maybe we can also just link internally/directly to the plugin install page rather than sending them off to jetpack.com? Although I guess the only way to do that is actually via a search, which is not great: `wp-admin/plugin-install.php?s=jetpack&tab=search&type=term`

donnchawp: plugin-install.php will redirect to wp-admin/network/plugin-install.php on an MS install, but Jetpack is one of the featured plugins so that isn't too bad.

donnchawp: I updated the code with a link to plugin-install.php. I didn't change the second sentence though because Jetpack is mentioned so much in the previous text.

PR #645, opened 2019-01-17
Update origin pull link

bonny:
It was going to a 404, probably because SoftLayer was acquired by IBM and is now called/part of IBM Cloud.


donnchawp:
Oops, it's a 404 again! Maybe it should just link to https://en.wikipedia.org/wiki/Content_delivery_network instead :)

PR #657, opened 2019-03-09
Add wpsc_is_caching_user_disabled to #631

donnchawp:


donnchawp:
@stodorovic - I seem to have created a new pull request instead of modifying your branch in #631, sorry. This one makes the "do not cache known users" option into a series of radio boxes like this:

Screenshot 2019-03-09 at 16 31 25

donnchawp:
I added a cookie banner plugin that sets a cookie "euCookie" and added that to WP Super Cache with:
```
function add_wpsc_cookies() {
    do_action( 'wpsc_add_cookie', 'euCookie' );
}
add_action( 'init', 'add_wpsc_cookies' );
```

1. page was cached using a non-anonymous .php file
2. page was not cached or served from cache.
3. page was cached again.

When loading the page by a logged in user the page was only cached when option 1 was checked. :)

I noticed the exit I added to the phase2 function stopped the html comments being added to the page. I'll have to fix that.

donnchawp:
Added the html comment in that last check in.

stodorovic:
It seems fine for me. I'll try to test this PR in next days.

donnchawp:
In testing "disable for logged in users" the media uploader stopped displaying images. Needs more work. :(

stodorovic:
It's related to HTML comment. I see error: `JSON.parse: unexpected non-whitespace character after JSON data in javascript`.
It's an issue with `wpsc_shutdown_message`. I fixed it when I've changed this function:
```php
function wpsc_shutdown_message() {
        if ( wpsc_is_backend() || ! defined( 'WPSCSHUTDOWNMESSAGE' ) ) {
                return false;
        }
```
I think that's better to we move register_shutdown_function later - maybe before creating output buffers. I'll check more cases (feeds, sitemaps,...).

juanrules:
> @stodorovic - I seem to have created a new pull request instead of modifying your branch in #631, sorry. This one makes the "do not cache known users" option into a series of radio boxes like this:
> 
> Screenshot 2019-03-09 at 16 31 25


Just to note that after the above update, I started getting the below in all pages: 
`
`

It looks like it is because the advanced setting “Don’t cache pages for known users”  was  replaced with “Disable caching for visitors who have a cookie set in their browser”.  

To fix it, I had to set it the 3rd option "Only logged in visitors" as per my previous configuration. Now, given the set up of my site, I had to do it on the wp-cache-config.php file by setting the $wp_cache_not_logged_in  var to 2:

`$wp_cache_not_logged_in = 2;`

I hope that helps if anyone else comes across the same issue. 
P.S.: It looks like [it has been reported here too ](https://wordpress.org/support/topic/dont-cache-pages-for-known-users-2/)

PR #670, opened 2019-05-12
Fix updating wpcachehome

donnchawp:
wp_cache_replace_line() used to return false when a setting hadn't changed but also did that when a fatal error occurred. This patch changes it so it returns true if a setting hasn't changed.
This had been used when enabling or disabling the cache so I had to change those bits of code too.
However, it makes checking for fatal errors much easier now.


donnchawp:
This patch also removes the line that in errors says CACHEHOME has to be modified in advanced-cache.php as mentioned in https://wordpress.org/support/topic/instruction-to-modify-cachehome-is-not-clear/

PR #672, opened 2019-05-21
Added support for ONLY_SUPERCACHE and DONT_SERVE_CACHE_FILE constants

Webbmekanikern:
We just wanted to use the supercached, static HTML files, and got a lot of issues with the cached PHP files. By defining the constant `ONLY_SUPERCACHE` in `wp-config.php`, no cached PHP file is ever generated.

We also added support for totally ignoring any cached file by defining the constant `DONT_SERVE_CACHE_FILE`, as we are letting the web server do the serving of the supercached files. This saves us some unnecessary file lookups.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -77,7 +77,7 @@ function wp_cache_serve_cache_file() {
 			wp_cache_debug( 'Meta array from object cache corrupt. Ignoring cache.', 1 );
 			return true;
 		}
-	} elseif ( ( $cache_file && file_exists( $cache_file ) ) || file_exists( get_current_url_supercache_dir() . 'meta-' . $cache_filename ) ) {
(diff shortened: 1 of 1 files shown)


donnchawp:
In effect you're only serving cached files to anonymous users? Did you try checking "Don’t cache pages for known users. (Recommended)" on the advanced settings page? That should do the same thing.

Webbmekanikern:
We want to cache the pages to static files regardless of you're logged in or not. The content will always be the same.

donnchawp:
Try checking "Make known users anonymous so they're served supercached static files" on the advanced settings page. That will make them see logged out pages.

donnchawp:
This branch had conflicts because of recent changes so in https://github.com/Automattic/wp-super-cache/commit/d32360b7b2a4bf5413f911a9668359d07fdb263d I added WPSC_SERVE_DISABLED and WPSC_SUPERCACHE_ONLY. Thanks for the PR.

PR #701, opened 2019-07-26
Updated 'admin bar' verbiage to 'Toolbar' 

garrett-eclipse:
Found this while translating updating the verbiage to match the change in core WP.


donnchawp:
Thanks!

PR #709, opened 2019-09-19
Fix it's -> its and document wp-super-cache-plugins

raphj:


donnchawp:
Thanks @raphj  for the patch!

Issue #713, opened 2019-10-29
Need action after cache cleared (post publish, edit, delete cache, etc.)


janvitos:
Hi,

I'm looking for an action that is executed when the cache is cleared in any possible way (post publish, edit, delete cache clicked, etc.)

I found the wp_cache_cleared action, but that is only executed when the delete cache button is clicked.

I could not find anything in the code that does this, so it might be helpful to add a new action, perhaps inside the wp_cache_post_edit function.

Thank you.

PR #714, opened 2019-10-29
Add wp_cache_cleared action to prune_super_cache

janvitos:
The wp_cache_cleared action was missing for the prune_super_cache function. This enables a function to be hooked after the cache is cleared when a post is published / edited / trashed etc.

Fix for ticket #713


janvitos:
This won't work at prune_super_cache is called multiple times and calls the wp_cache_cleared action multiple times as well.

PR #716, opened 2019-11-05
Add filter to wp-config.php WP_CACHE setting; Close #715

### LINKED ISSUE #715: Add filter to change `WP_CACHE` line which gets written to wp-config.php
I need to entirely exclude IE from viewing cached pages. To do this, I can modify `define('WP_CACHE', true)` in my `wp-config.php` to the following:

```php
if (!(stripos($_SERVER["HTTP_USER_AGENT"], "MSIE") || stripos($_SERVER["HTTP_USER_AGENT"], "Trident"))) {
    define('WP_CACHE', true);
}
```

This works perfectly, but I'd like to be able to change this line via a filter in my theme. I see that the line gets defined at `wp-cache.php:2665`. Can we please get a filter on this line so as to facilitate modification?

JacobDB:
This is a simple solution for #715. It allows developers to customize the line that gets written to `wp-config.php`, thus allow for conditional caching (for example, disabling caching when using IE).

I tested this, and it gets added and removed correctly even when customized with a filter in a theme, so I believe this should work without any consequences, but please do let me know if I'm missing something obvious.

### DIFF
--- a/wp-cache.php
+++ b/wp-cache.php
@@ -2662,7 +2662,7 @@ function wp_cache_check_global_config() {
 		$global_config_file = dirname( ABSPATH ) . '/wp-config.php';
 	}
 
-	$line = 'define(\'WP_CACHE\', true);';
+	$line = apply_filters('supercache_wp_config_line', 'define(\'WP_CACHE\', true);');
 	if (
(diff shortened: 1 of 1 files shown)


donnchawp:
Not merging as there are other ways to do conditional caching.

PR #717, opened 2019-11-16
Rename readme.txt to readme.md

j5ingh:
Github 'renders' _.md_ files on the repo overview page so it looks nicer.


donnchawp:
We can't rename the readme because readme.txt is used by WordPress.org but I should create a short readme.md summary linking to the plugins page.

j5ingh:
Okay. Happy to help if needed.

PR #719, opened 2019-12-06
Improve wording of file list headers

jakejarvis:
Hello! A very simple PR here.

The file lists under the Contents tab had headings that ended in 2x the necessary "Files":

Screen Shot 2019-12-05 at 8 56 02 PM

I removed the word from both strings since "Files" already being appended in line 2886.

While I was there, I changed the wording to match the two categories right above it, from "Super Cached" to "WP-Super-Cached" and "Full Cached" to "WP-Cached". I can revert those to the old wording if needed.

Thanks!


donnchawp:
Thanks @jakejarvis for the patch!

PR #720, opened 2019-12-16
Add static brotli compression support.

stathis:
To be used in conjunction with https://github.com/kjdev/php-ext-brotli


vlasky:
@donnchawp @stathis 's code that adds Brotli support seems straight forward.

Brotli does a lot to further reduce bandwidth usage beyond what can be achieved with gzip. Can you please merge this commit?

To this day, I have been running a shell script to perform brotli compression of static assets, served by Brotli-enabled Nginx with the brotli_static configuration option enabled.

It works, but it would be so much more elegant and convenient to have this done by WP Super Cache.

PR #722, opened 2020-01-07
fix(check_up_on_preloading()): Adding array type check before using option value.

### LINKED ISSUE #721: check_up_on_preloading() - notice: Trying to access array offset on value of type bool - wp-super-cache/wp-cache.php:3716

spenserhale:
Closes #721


donnchawp:
Thanks, I was going through the forum posts and added the fix in #724.

PR #730, opened 2020-02-15
Fix plugin for usage with Debian wordpress package

smortex:
Debian packages wordpress in a way that a single wordpress instance can
be used to serve multiple websites.  The wp-config.php file is stored
in a well-known location: /etc/wordpress/config-$HOST.php as explained
in the [README.Debian] file.

The package defines DEBIAN_FILE as the filename of the configuration
file.  When it is defined, and if the wp-config.php was not found using
the "traditional" way, use it.

By default, this file is not writable by the application, adding the
following to it is enough to use the plugin:

```
define('WP_CACHE', true);
define('WPCACHEHOME', '/var/lib/wordpress/wp-content-xxxxx/plugins/wp-super-cache/');
```

[README.Debian]:https://salsa.debian.org/debian/wordpress/blob/master/debian/README.debian


donnchawp:
Thanks for the patch. I fixed some formatting but it looks good.

PR #747, opened 2020-10-29
Rewrite the plugin

donnchawp:
Initial import of framework for rewrite.
Also includes some work on configuration saving.


CGastrell:
Since you're working almost "solo" on this, I'd just approve it. But if you come up with some simple tests so I can say "works", I'd have no problem testing it :)

stathis:
While you are at it, why don't you add php-brotli support to it?
I'm already maintaining a fork of WPSC with brotli support, maybe it could be of help.

[This extension](https://github.com/kjdev/php-ext-brotli) is already in some repositories, and most Nginx nowadays support `brotli_static` .

PR #753, opened 2021-02-01
PHPCBF Everything

kraftbj:
Adding the rules proposed in #752, here is what PHPCBF could automatically fix.﻿


kraftbj:
(I don't expect this to merge. I figure get the rewrite figured out and landed first, but opening this just to have something to look at)

donnchawp:
Nice. I should have run this years ago ...

lkraav:
:rocket: 

While y'all at this, it's worth it to also look into auto-linting pre-commit hook setup.

We do it like https://github.com/conversionxl/lint-staged-config/blob/ea0b9a4b3696abd67b40809ec77974ed4b5f8e32/index.js#L6-L28 for auto-running `phpcbf`, `phpcs`, `psalm` on all to-be-committed files locally.

kraftbj:
@lkraav I concur. [This is what we use over in Jetpack](https://github.com/Automattic/jetpack/blob/master/tools/git-hooks/pre-commit-hook.js) that does PHPCS and eslint, with phpcs-changed to force changed lines to be compliant and we made our own variant of it to work with eslint for JS.

kraftbj:
Going to close this to defer to #754. Merging this while there's a rewrite in another branch is just going to be asking for merge headaches later.

PR #754, opened 2021-02-02
v2 rewrite: PHPCBF

kraftbj:
Like Automattic/wp-super-cache#753, but into the Automattic/wp-super-cache#747 v2_rewrite branch.


donnchawp:
My eyes are going square from looking over this. It's large!
I still need to look over wp-cache-phase2.php and wp-cache.php but TBH as it's going into the rewrite branch which doesn't do much anyway so I would be happy to merge it if you are.

kraftbj:
I hear ya. I defer to your judgement. The autofixes are almost always fine (I can't think of a time that one actually broke something).

The manual fixes, sure. I'm good with merging.

nekohayo:
Hi there, I sadly can't help with this technically but as an enthusiast I'm wondering what's the status and relationship of this PR and PR 747. This appears to be stalled but seemed close to completion? The current mainline/stable version has so many issues that prevent me from having working caching (like Automattic/wp-super-cache#864, or HTTPS or preload/GC-related heisenbugs, to the point where I don't know what is what anymore) that I'm secretly hoping that this merge request / branch might solve many things in one sweep... I'm just touching wood, expressing my cautious optimism and hoping that this effort has not been abandoned!

PR #755, opened 2021-02-04
Make config file path configurable

sebastianpopp:
By default the `wp-cache-config.php` file is saved in the `wp-content` directory and deleted and newly created on every config change. That causes problems if the `wp-cache-config.php` file is a symlink (see Automattic/wp-super-cache#884). To create a better workaround, I added a constant `WPCACHECONFIGPATH`. Now one can set
```php
define( 'WPCACHECONFIGPATH', '/var/www/current/wp-content/cache-config' );
```
in `wp-config.php` and make the whole `cache-config` directory a symlink.
Since `WPCACHECONFIGPATH` is set to `WP_CONTENT_DIR` if it's not defined, there's no change in behavior unless explicitly set.

### DIFF
--- a/wp-cache-phase1.php
+++ b/wp-cache-phase1.php
@@ -5,7 +5,11 @@ if ( ! function_exists( 'wp_cache_phase2' ) ) {
 }
 
 // error_reporting(E_ERROR | E_PARSE); // uncomment to debug this file!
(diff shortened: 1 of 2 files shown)


donnchawp:
Looks good! I'll need to modify a few more messages, and add a message to the advanced settings page about this but this is useful! Thanks!

PR #766, opened 2021-04-28
Fix example URL

hirasso:
Under "Directly Cached Files", the example URLs where using
`siteurl`, which lead to wrong URLs in WP installs with a custom folder structure.

Using 'home' fixes this minor issue.


donnchawp:
Thanks for that fix!

Issue #712, opened 2019-10-07
Ignore query string, not bypass cache


kristos80:
Hi there,

Thanks for your great plugin first of all.

I found a lot of support tickets, blog comments, etc, regarding a huge issue --in my opinion, that doesn't seem to have been issued with a permanent solution: The ability of the plugin to ignore specific query string parameters.

With huge traffic coming nowadays from social networks and/or 3rd party sources, any caching mechanism should be able to ignore specific query string parameters (such as utm sources) without the need for URL rewriting.

So this is not an issue per se, but rather a request for a feature that is extremely important.

astbarc:
I totally agree with you, this is a very huge problem. Any visit from facebook or google ads is never cached.

Does anyone know any way to implement this? Any workarounds?

idoschor:
@astbarc Overriding `$wp_cache_request_uri` has worked for me. As suggested here: https://wordpress.org/support/topic/ignore-parameters/

mahnunchik:
I think it should be added to documentation. Also interesting how to configure apache to skip `utm_*`, `gclid` etc params.

astbarc:
Thanks for solution @idoschor 

@donnchawp  Would be great for the plugin to have an option to indicate which parameters we want to exclude: utm, gclid, fbclid instead of having to hardcode the plugin.

tschoartschi:
Is it still the best option, to modify the plugin? I guess this could be troublesome when I want to upgrade the plugin 🤔

lkraav:
Correct, https://github.com/Automattic/wp-super-cache/issues/712#issuecomment-547336896 is at best a temporary hack solution.

Plugin should provide filter hooks or configuration options for this. Before Comet Cache died, we were working on defining implementation requirements at https://github.com/wpsharks/comet-cache/issues/850

There's more to consider than meets the eye, incl. always normalizing query param order to determine correct caching status.

But definitely doable, if someone invested time and/or money.

This could potentially be a solid Upwork target.

lukapaunovic:
GCLID, FBCLID, IGSHD....
and many others, this is a huge problem.

lynt-smitka:
@mahnunchik for serverside skip you can try our configs:

Apache (.htaccess): https://gist.github.com/lynt-smitka/297004bed3aa4fb3065af16fd6e7ff18
Nginx (config): https://gist.github.com/lynt-smitka/b8bf0a4b7e2a9e5802faa48ee16b19b0

[...]

PR #770, opened 2021-05-10
Ignore tracking parameters

radex02:
Filter out tracking parameters (from Facebook, Google, Matomo) to use cached page as if there were no tracking parameter.
Pretty much a copy/paste of what @markfinst suggested in his comment [here](https://github.com/Automattic/wp-super-cache/issues/712#issuecomment-814626300)

Temporary fix for #712

Optimal solution would be to have a user friendly way of editing the ignored parameters

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -46,7 +46,7 @@ function remove_tracking_params_from_uri($uri) {
 	$path = isset($parsedUrl['path']) ? $parsedUrl['path'] : '';
 	$query = !empty($query) ? '?'. http_build_query($query) : '';
 
-	return $parsedUrl['scheme']. '://'. $parsedUrl['host']. $path. $query;
+	return $parsedUrl['host']. $path. $query;
 }
 
 function wp_super_cache_init() {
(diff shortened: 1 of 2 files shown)


wp-cache-phase2.php
```diff
@@ -28,6 +28,27 @@ function get_wp_cache_key( $url = false ) {
...
+			'mtm_source', 'mtm_medium', 'mtm_campaign', 'mtm_keyword', 'mtm_content', 'mtm_cid', 'mtm_group', 'mtm_placement',
+		);
+		foreach($q_params_to_ignore as $param_name) {
+			unset($query[$param_name]);
+		}
+	}
+	$path = isset($parsedUrl['path']) ? $parsedUrl['path'] : '';
+	$query = !empty($query) ? '?'. http_build_query($query) : '';
+
+	return $parsedUrl['scheme']. '://'. $parsedUrl['host']. $path. $query;
```
mikaelz, line 49: I tried this, but had issue with undefined scheme and host. My $uri was just `/some-name/`

donnchawp:
I updated this PR and added a setting in #777 if you want to give it a go.

donnchawp:
Thanks for the PR. Added this functionality in #777.

PR #797, opened 2021-12-10
Check $cache_path is somewhat sane.

donnchawp:


donnchawp:
Ref: https://wordpress.org/support/topic/open_basedir-restriction-5/

PR #829, opened 2022-07-05
Replaced dirname(__FILE__) with __DIR__

AntarMukhopadhyaya:
Replace the dirname(__FILE__) with __DIR__ to make the code shorter


kraftbj:
I'm afraid we can't use this yet. WP Super Cache still supports older versions of WP which support PHP 5.2.

The DIR magic constant was introduced in 5.3.

kraftbj:
Suggesting bumping the minimum version in #839.

PR #833, opened 2022-07-19
Adding github action for deploy to WP.org repo

mikestottuk:
### Description

This is a developer tooling PR which adds an action to automate the deploy to WP.org when a new release is **published**

### Testing instructions

When a new release is published - the github action will run and deploy to WP.org.


kraftbj:
I'd suggest dropping this for now. We have some automation basically ready, but not yet implemented, in the monorepo to handle this.

Going to close, but won't delete the branch in case y'all disagree.

PR #839, opened 2022-08-12
Bump minimum supported WP to 5.2 and PHP to 5.6

kraftbj:
I think it's okay to start bumping this up to more modern versions. WP 5.2 is the first one that dropped PHP 5.2 support, so proposing we start there.


kraftbj:
One idea is to bump up to latest WP stable minus 1 and stick with that going forward, which matches what Automattic does with the Jetpack ecosystem. cc: @thingalon

thingalon (approved):
I agree. PHP 5.6 is a good base-line and more consistent with other Automattic products.

PR #843, opened 2023-01-04
Fix wp-super-cache changing file permissions after updating it

### LINKED ISSUE #27324: [Super Cache] wp-config permissions changed on activation / deactivation
Relevant link: https://wordpress.org/support/topic/wp-config-php-permissions-set-to-600/

A user has observed that their wp-config.php file had its permissions changed from 644 to 600 during installation and uninstallation of Super Cache.

Super Cache updates wp-config.php to add the `WP_CACHE` and `WPCACHEHOME` constants to facilitate managing the cache before WordPress (and the connection to its database) has finished loading. To do so safely, Super Cache copies wp-config.php to a new file with the relevant edits - then renames the new file in place of the old one. Along the way, it doesn't copy the file permissions from the original wp-config.php to the new one. 

We should add a chmod to the process to ensure the updated wp-config.php file has the same permissions as the original.

dilirity:
#### Description

Fixes https://github.com/Automattic/jetpack/issues/27324.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -1400,7 +1400,9 @@ function wp_cache_replace_line( $old, $new, $my_file ) {
 		}
 	}
 	fclose( $fd );
+	$my_file_permissions = fileperms( $my_file );
 	rename( $tmp_config_filename, $my_file );
+	chmod( $my_file, $my_file_permissions );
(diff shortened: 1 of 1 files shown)


dilirity:
Oops - https://github.com/Automattic/jetpack/tree/trunk/projects/plugins/super-cache.

PR #1021, opened 2026-04-07
Fix: hide debug HTML comments in REST and Ajax responses

kraftbj:
## Summary

- Prevents WP Super Cache from appending HTML debug comments (`<!-- ... -->`) to non-HTML responses such as REST API, Ajax, JSON API, and WooCommerce API requests.
- Adds checks for `REST_REQUEST`, `JSON_REQUEST`, `WC_API_REQUEST` constants and `wp_doing_ajax()` to both `wp_cache_append_tag()` and `wp_cache_add_to_buffer()`, bailing early when the response is not standard HTML.

Originally proposed by @jom in #436.

This is a manual cherry-pick of the changes from closed PR #436, adapted to the current codebase. See #436.

## Test plan

- [ ] Verify REST API responses (e.g., `/wp-json/wp/v2/posts`) do not contain `<!-- ... -->` debug comments when Super Cache debug mode is enabled.
- [ ] Verify Ajax responses (`admin-ajax.php`) do not contain HTML debug comments.
- [ ] Verify WooCommerce API responses do not contain HTML debug comments.
- [ ] Verify normal HTML page responses still include debug comments when debug mode is enabled.
- [ ] Verify that non-HTML responses (no `<html` tag) continue to be excluded as before.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -2196,6 +2196,7 @@ function wpsc_skip_debug_output( $buffer ) {
 		( defined( 'REST_REQUEST' ) && REST_REQUEST ) ||
 		( defined( 'JSON_REQUEST' ) && JSON_REQUEST ) ||
 		( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) ||
+		( defined( 'XMLRPC_REQUEST' ) && XMLRPC_REQUEST ) ||
 		( function_exists( 'wp_doing_ajax' ) && wp_doing_ajax() );
 }
 
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -2210,7 +2210,14 @@ function wp_cache_append_tag( &$buffer ) {
 		$msg = "Live page served on $timestamp";
 	}
 
-	if ( strpos( $buffer, '<html' ) === false ) {
+	// Don't output debug message if we aren't handling a normal HTML response.
+	if (
+		strpos( $buffer, '<html' ) === false ||
+		( defined( 'REST_REQUEST' ) && REST_REQUEST ) ||
+		( defined( 'JSON_REQUEST' ) && JSON_REQUEST ) ||
+		( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) ||
+		( function_exists( 'wp_doing_ajax' ) && wp_doing_ajax() )
+	) {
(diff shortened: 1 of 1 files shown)


donnchawp:
> *This comment was generated by AI.*

Pushed a small refactor on top of your cherry-pick (2ac3825) and marked the PR ready for review.

The REST/Ajax/JSON/WC bail condition was copy-pasted identically into both `wp_cache_append_tag()` and `wp_cache_add_to_buffer()`. I moved it into a single predicate so the two sites can't drift:

```php
function wpsc_skip_debug_output( $buffer ) {
	return strpos( $buffer, '<html' ) === false ||
		( defined( 'REST_REQUEST' ) && REST_REQUEST ) ||
		( defined( 'JSON_REQUEST' ) && JSON_REQUEST ) ||
		( defined( 'WC_API_REQUEST' ) && WC_API_REQUEST ) ||
		( function_exists( 'wp_doing_ajax' ) && wp_doing_ajax() );
}
```

Both call sites are now `if ( wpsc_skip_debug_output( $buffer ) ) { ... }`, each keeping its own `wp_cache_debug()` line and `return false`. No behavioural change — same constants and guards as before; the named function also let the explanatory comment go. Net −6 lines.

One open item: this path has no automated coverage (see #1051), so the suppression is currently verified by the manual test plan only.

donnchawp:
> *This comment was generated by AI.*

Note on the `XMLRPC_REQUEST` clause I added to `wpsc_skip_debug_output()` (737ffb9): it's defensive/documentation, not load-bearing.

In practice XML-RPC requests never reach this code path — `xmlrpc.php` is already classified as backend in `wpsc_is_backend()` (`wp-cache-phase2.php:803`) and is in the `$auto_rejected` list (`wp-cache-phase2.php:1877`), so the output buffer is never built for it. `XMLRPC_REQUEST` is only defined inside `xmlrpc.php`, so the clause can't fire unless some other page both defines that constant and pushes XML through the cache buffer — a case that doesn't occur today.

It's kept for intent/parity with `advanced-cache.php` (which also enumerates `XMLRPC_REQUEST`), not because it's reachable.

The other clauses **are** reachable and necessary: REST (`/wp-json/`) rewrites to `index.php` so it isn't rejected upstream, and a JSON/REST body can contain the literal substring `<html`, which would defeat the old `strpos` sniff on its own — that's the actual bug this PR fixes.

PR #1023, opened 2026-04-07
Fix: use fileperms() instead of stat() and fix escaping

kraftbj:
## Summary

- Replace `stat()` with `fileperms()` for retrieving directory permissions in `wp-cache.php`. `fileperms()` returns the mode directly, removing the need for array access on the `stat()` result.
- Fix escaping in the mobile rewrite rules warning block: use `esc_html_e()` for plain-text translatable strings and `esc_html()` for variables passed to `printf()`.
- Fix an unclosed `<p>` tag in the same block.

Originally proposed by @stodorovic in #607.

See #607


donnchawp (approved):
Looks good! Thumbs up from me!

PR #1026, opened 2026-04-07
Add specific time scheduling to Preload

### LINKED ISSUE #1017: Enhancement:Time specific preload in wp-super cache
### Impacted plugin

Super Cache

### What

Chose a specific time to preload the cache instead of time intervals 

### How

Hi, is there a way to set the preload option to a specific time as there is on the garbage collection instead of just time intervals? I think it would be a good option to schedule the preload with other automated tasks on the server to be in the most optimal order and with the least traffic on the site.

kraftbj:
## Summary

- Add a "Scheduler" option to the Preload settings with "Timer" (interval-based) and "Clock" (specific time) modes
- Port the "Clock" UI selector from Garbage Collection settings to the Preload tab for consistency
- Update scheduling logic in `wp_cron_preload_cache()` and `wpsc_preload_settings()` to handle both interval and time-based scheduling

This allows server administrators to schedule resource-intensive preloading during off-peak hours (e.g., 03:00 AM) to minimize impact on site performance.

Originally proposed by @zeus2611 in Automattic/jetpack#47018.
Fixes #1017

## Test plan

- [ ] Navigate to Settings > WP Super Cache > Preload tab
- [ ] Verify a "Scheduler" section with "Timer" and "Clock" radio buttons is present
- [ ] Test Clock mode: select Clock, enter a time, save, and verify `wp_cache_preload_hook` is scheduled at the correct time (e.g., via WP Crontrol)
- [ ] Test Timer mode: select Timer, enter an interval, save, and confirm the hook is scheduled based on the interval
- [ ] Verify switching between modes reschedules correctly


donnchawp:

Reviewed the scheduling state machine, since converting `wp_cache_full_preload_hook` from a self-rescheduling single event to a recurring event in "time" mode is the risky part. Findings below; I verified the load-bearing ones against the code.

## Context

Two cron hooks both fire `wp_cron_preload_cache` (`wp-cache.php:3748-3749`): `wp_cache_preload_hook` (per-batch loop) and `wp_cache_full_preload_hook` (full-preload kickoff). The PR makes the latter recurring in "time" mode but only updates some of the paths that manage it.

## Blocker

**`wpsc_cancel_preload()` cannot cancel a recurring time-mode preload** — `wp-cache.php:3990`

```php
wp_unschedule_event( $next_full_preload, 'wp_cache_full_preload_hook' );
```

`wp_unschedule_event()` removes only the *next* occurrence. For the recurring event that "time" mode now creates, all future occurrences survive — so clicking "Cancel Cache Preload" leaves preloading to resume at the next wall-clock time. `wpsc_cancel_preload()` isn't touched by this PR, even though `wp_clear_scheduled_hook()` *was* added to the reschedule path in `wpsc_preload_settings()`. Same fix needed here:

```php
wp_clear_scheduled_hook( 'wp_cache_full_preload_hook' ); // handles single + recurring
```

The `preload_off` flow routes through `wpsc_cancel_preload()`, so fixing this one function covers both entry points.

## Cleanup (not a blocker)

**End-of-run reschedule block in "time" mode is dead code** — in `wp_cron_preload_cache()`'s else-branch, the guard `! wp_next_scheduled( 'wp_cache_full_preload_hook' )` is always false while the recurring event is alive, so the `wp_schedule_event()` there never runs. No double-scheduling occurs (the guard holds), but the block is misleading. In time mode the recurring event self-fires — do nothing at end-of-run and drop that `wp_schedule_event()` call.

## Warnings

- **`.attr('checked', true)` → `.prop('checked', true)`** in the new inline script (`partials/preload.php`). In jQuery 1.6+, `.attr` doesn't reliably toggle the radio's checked property, so clicking the time field or interval dropdown can silently fail to select its radio.
- **Missing `phpcs:ignore WordPress.Security.NonceVerification.Missing`** on the `preload_scheduled_time` `$_POST` block, while its two siblings have it — this will fail the changed-lines PHPCS lint in CI.
- **Mode switch mid-preload** (interval → time) leaves orphaned `wp_cache_preload_hook` loop events; the reschedule path only clears `wp_cache_full_preload_hook`.
- **`wpsc_enable_preload()` ("Preload Now") is not mode-aware** — it always fires a single event. It currently survives via the `wp_next_scheduled` guard, but it's fragile alongside the recurring registration.

## Not issues

Timezone handling in `wpsc_next_scheduled_preload_timestamp()` is sound — `createFromFormat('H:i')` inheriting today's date is intended, and `modify('+1 day')` is DST-safe. The regex-then-`substr` fallback is unreachable but harmless.

## Suggestion

The root cause is that "convert to recurring" wasn't applied uniformly across the schedule / cancel / enable paths. Consider centralizing into a single `wpsc_schedule_preload()` / `wpsc_unschedule_preload()` pair that both modes route through, so these paths can't diverge again.

PR #1043, opened 2026-04-25
Fix: add null check for comment in wp_cache_get_postid_from_comment()

### LINKED ISSUE #1014: PHP Warning in wp-cache-phase2.php: Array Offset on Null
### Impacted plugin

Super Cache

### Quick summary

On multiple of our sites, we’re seeing this warning :

```text
PHP Warning:  Trying to access array offset on null in ***/plugins/wp-super-cache/wp-cache-phase2.php on line 3024
```

The relevant code is inside the `wp_cache_get_postid_from_comment()` function in `wp-cache-phase2.php`, specifically this line:

```php
$comment['old_comment_approved'] = $comment['comment_approved'];
```

The surrounding snippet is:

```php
$comment = get_comment( $comment_id, ARRAY_A );
if ( $status != 'NA' ) {
    $comment['old_comment_approved'] = $comment['comment_approved'];
    $comment['comment_approved']     = $status;
}
```

It appears that `get_comment( $comment_id, ARRAY_A )` is returning `null`, so attempting to access `$comment['comment_approved']` triggers the warning. We suspect this happens when the cache‐clear routine runs after a comment has already been deleted.

Would it make sense to add a null‐check for `$comment`, for example:

```php
if ( ! $comment ) {
    wp_cache_debug( "Comment not found; skipping cache update.", 4 );
    return;
}
```

before trying to access its array offsets?

Environment when this error occurred:
- PHP 8.3.8
- WordPress 6.7
- WP Super Cache version 2.0.0

### Steps to reproduce

I attempted to reproduce the error but wasn’t able to.

This warning may have occurred when deleting a comment with the command (using the --force option)
```
wp comment delete [comment id] --force
```
 rather than through the admin dashboard.

### Site owner impact

Fewer than 20% of the total website/platform users

### Severity

Minor

### What other impact(s) does this issue have?

No revenue impact

### If a workaround is available, please outline it here.

_No response_

### Platform (Simple and/or Atomic)

Simple

DeveloperBatuhanALGUL:
Fixes #1014

get_comment() returns null when a comment has already been deleted by
the time the cache-clear routine runs. The existing code attempts to
access array offsets on that null value immediately after, which
triggers a PHP Warning on PHP 8.x.

This adds an early return right after get_comment() so the function
exits cleanly when the comment no longer exists, rather than
proceeding with a null value and generating noise in the error log.

Tested on PHP 8.3.8 / WordPress 6.7 / WP Super Cache 3.1.0.

### DIFF
--- a/wp-cache-phase2.php
+++ b/wp-cache-phase2.php
@@ -3014,6 +3014,10 @@ function wp_cache_get_postid_from_comment( $comment_id, $status = 'NA' ) {
 	}
 
 	$comment = get_comment( $comment_id, ARRAY_A );
+	if ( ! $comment ) {
+		wp_cache_debug( "Comment not found; skipping cache update.", 4 );
+		return;
+	}
 	if ( $status != 'NA' ) {
 		$comment['old_comment_approved'] = $comment['comment_approved'];
 		$comment['comment_approved']     = $status;


donnchawp (approved):
Thanks for this — the fix correctly resolves the warning in #1014.

`get_comment()` only returns `null` here when the comment row is genuinely gone from the DB, and at that point there's no recoverable `comment_post_ID` to target a precise cache clear with. The previous fall-through to `wp_cache_post_change( wp_cache_post_id() )` resolved to a no-op in the admin moderation context anyway (`wp_cache_post_id()` returns 0 there → `get_post(0)` bails), so the early return is equivalent in the common case and additionally silences the warning. Good, minimal change.

For the record: the only thing the early return forfeits is the generic "clear the currently-rendered page" fallback, which is meaningful only in a front-end request context that the null case rarely coincides with. A strictly-more-conservative alternative would be to guard the array dereferences and let the function fall through to the existing clear, but when the comment is truly deleted that mostly resolves to the same no-op — not worth the extra surface area here.

Approving as-is.

Issue #1051, opened 2026-05-27
Make procedural cache functions unit-testable and seed initial PHPUnit tests


donnchawp:
> *This issue was generated by AI.*

## Context

The PHPUnit scaffolding already exists and runs in CI, but exercises nothing:

- Configs `phpunit.{9,11,12}.xml.dist`, composer scripts (`test-php`/`phpunit` via `phpunit-select-config`), `yoast/phpunit-polyfills`, and `tests/php/bootstrap.php` are all present.
- `.github/workflows/php-tests.yml` runs `composer test-php` on every PR/push across PHP 8.2–8.5.
- **Gap 1:** `tests/php/bootstrap.php` only loads the Composer autoloader. There is no WordPress runtime, and the core logic lives in procedural files (`wp-cache-phase1.php`, `wp-cache-phase2.php`) that are *not* autoloaded (the classmap covers `src/` only). So functions like `supercache_filename()` are unreachable from a test.
- **Gap 2:** Zero `*Test.php` files exist, so the CI step passes trivially and gives false assurance.

## Goal

Make the procedural caching functions testable and land a first batch of tests, so the existing green CI step actually means something.

## Decision required (why this is human, not agent)

Pick a testing strategy — this is an architectural call with trade-offs:

1. **WordPress integration test library** (wp-env / `wordpress-tests-lib`) — real WP + DB. Highest fidelity, slowest, needs DB service in CI.
2. **Function mocking** (Brain Monkey / Mockery) — mock `apply_filters`, `get_blog_option`, etc. Fast, no DB, but mocks can drift from real WP behaviour.
3. **Lightweight stubs in bootstrap** — define the handful of WP functions the targeted code calls. Minimal deps; only viable while the surface stays small.

Whichever is chosen, the bootstrap must also `require` the procedural files (or the relevant ones) so their functions are defined.

## First test targets (pure / global-light)

- `supercache_filename()` — assert the output is always a single, separator-free path segment given hostile filter input (regression guard for the filename-sanitization fix in #1050).
- `wpsc_get_realpath()` / `wpsc_is_in_cache_directory()` — path containment logic.
- `get_wp_cache_key()` — key composition.

## Acceptance

- `composer test-php` runs at least one real test green across the CI PHP matrix.
- Bootstrap loads the procedural files under the chosen strategy.
- A documented pattern (README or `tests/php` note) so future tests have a template to copy.

donnchawp:
> *This was generated by AI during triage.*

[...]

PR #1057, opened 2026-05-29
Fix: parse Accept header q-values in wpsc_get_accept_header()

### LINKED ISSUE #1045: wpsc_get_accept_header() ignores Accept header quality values (q-values)
`wpsc_get_accept_header()` in `wp-cache-phase2.php` uses a naive `str_contains()` check to detect JSON requests. It does not parse Accept header quality values (`q=`) as defined in [RFC 7231 §5.3.2](https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2). As a result, any request where `application/json` appears anywhere in the Accept header — even at low priority — is misclassified as a JSON request, preventing the cached file from being served and triggering a cache rebuild.

This is reproducible with New Relic Synthetics, which sends:

```
Accept: text/html,application/xhtml+xml,application/json;q=0.9,application/javascript;q=0.9,text/javascript;q=0.9,application/xml;q=0.9,text/plain;q=0.8,*/*;q=0.7
```

This is a valid HTML-preferring Accept header —` text/html` has implicit `q=1.0` (highest priority), and `application/json` is explicitly deprioritised at `q=0.9`. However, WPSC treats the entire request as a JSON request and refuses to serve the cached file.

---

In `wp-cache-phase2.php`, `wpsc_get_accept_header()` (around line 533):
```php
$accept = isset( $_SERVER['HTTP_ACCEPT'] ) ? strtolower( filter_var( $_SERVER['HTTP_ACCEPT'] ) ) : '';

foreach ( $accept_headers as $header ) {
    if ( str_contains( $accept, $header ) ) {
        $accept = 'application/json';
    }
}

if ( $accept !== 'application/json' ) {
    $accept = 'text/html';
}
```

The `str_contains()` check finds `application/json` anywhere in the raw Accept string without considering q-values. A request from New Relic Synthetics with `application/json;q=0.9` (lower priority than text/html) is classified as a JSON request.

This misclassification has two downstream effects:

1. `wp_cache_serve_cache_file()` returns `false` immediately because `wpsc_get_accept_header() !== 'text/html'`, so the cached file is never served.
2. `get_wp_cache_key()` appends `-application/json` to the cache key, meaning NR Synthetics requests generate and populate a separate cache bucket, triggering a fresh page build on every synthetic check.

---

### Steps to reproduce
1. Enable WP Super Cache with Expert/SuperCache mode
2. Enabled Debug logging 
3. Create a postman request that uses the above accept header
4. Observe in the debug log: `wp_cache_serve_cache_file: visitor does not accept text/html. Not serving cached file. repeating on every synthetic check`
5. Observe cache files being rebuilt on every NR Synthetics request

---

### Expected Bahviour
When `text/html` is the highest q-value type in the Accept header, WPSC should classify the request as `text/html` and serve the cached file normally. The function should parse q-values and determine the highest-priority accepted type before comparing against the known JSON types.

---

### Workarounds

The `wpsc_accept_headers` filter cannot easily fix this because `wpsc_get_accept_header()` caches its result via `static $accept = 'N/A'` and is called very early in the PHP request lifecycle — before `functions.php` is loaded — so any filter hooked from a theme or standard plugin loads too late.

A short-term workaround for those using WP Super Cache in PHP mode (not .htaccess/Expert mode) is to add via an **mu-plugin**:

```PHP
<?php
// mu-plugins/fix-nr-accept-header.php
add_filter( 'wpsc_accept_headers', function( $headers ) {
    if ( ! empty( $_SERVER['HTTP_X_NEWRELIC_SYNTHETICS'] ) ) {
        return [];
    }
    return $headers;
} );
```

However this is NR-specific and does not fix the underlying q-value parsing issue that would affect other monitoring tools (Datadog Synthetics, Pingdom, UptimeRobot, etc.) with similar broad Accept headers.

---

### Suggested Fix
`wpsc_get_accept_header()` should parse the Accept header properly, sorting types by q-value before comparing.

thisismyurl:
## Summary

Fixes #1045 — `wpsc_get_accept_header()` was using a naive `str_contains()` check that matched any JSON media type anywhere in the Accept string, ignoring RFC 7231 §5.3.2 quality values. The result: monitoring tools like New Relic Synthetics (and Datadog/Pingdom/UptimeRobot) that send valid HTML-preferring headers such as:

Accept: text/html,application/xhtml+xml,application/json;q=0.9,...

were misclassified as JSON requests. Two downstream effects:
1. `wp_cache_serve_cache_file()` refused to serve the cached file on every synthetic check.
2. A separate `application/json` cache bucket was populated on each check, triggering a fresh page build every time.

### Changes

**`wp-cache-phase2.php`**

- Extracted a new `wpsc_parse_accept_header( $raw_accept, $json_types )` helper. It parses q-values per RFC 7231 §5.3.2 and classifies the request as `application/json` **only when a known JSON type has a strictly higher q-value than `text/html`**. Ties resolve to `text/html` (safe default: serve the cached page).
- `wpsc_get_accept_header()` now delegates to this helper after the empty-header early-return. The static memoization and `wpsc_accept_headers` filter contract are unchanged. Callers are unaffected.

Edge cases handled:
- Malformed q-values (non-numeric, out-of-range) — treated as `q=1.0`, no warning or fatal.
- `*/*` wildcard — provides effective html q-value only when `text/html` is not explicitly present.
- Whitespace around media-range tokens.

**`tests/php/WpscParseAcceptHeaderTest.php`** (new file)

14 unit tests covering the seven acceptance criteria from the issue brief plus additional edge cases (wildcard behaviour, custom types via the filter, typical browser headers, out-of-range q-values). The test exercises `wpsc_parse_accept_header()` directly — it has no WordPress dependencies, so no WP test scaffolding is required.

PHPUnit 12.5.28 — 14 / 14 (100%) — 0.003s

> Note on test harness: the issue brief mentions a harness being stood up in #1051. I've added the test class here so the coverage lands with the fix rather than waiting. If #1051 introduces a shared bootstrap or test base class, this file is easy to migrate.

### Acceptance criteria from the issue brief

- [x] New Relic Synthetics header (`text/html` implicit q=1.0, `application/json;q=0.9`) → `text/html`
- [x] `Accept: application/json` alone → `application/json`
- [x] `Accept: text/html,application/json` (tie, both q=1.0) → `text/html`
- [x] `Accept: application/json,text/html;q=0.9` (json strictly higher) → `application/json`
- [x] `Accept: application/ld+json;q=1.0,text/html;q=0.8` (extended type via filter) → `application/json`
- [x] Absent or empty Accept → `text/html`
- [x] Malformed q-values → no warning or fatal, deterministic result
- [x] `wpsc_accept_headers` filter types participate in q-value comparison
- [x] Function remains free of late-loaded dependencies; static memoization preserved

### Release Notes

- Fix: honour quality values in the Accept header, so monitoring tools and browsers that list JSON below HTML are served the cached page instead of rebuilding it on every request.

### DIFF
--- a/tests/php/WpscParseAcceptHeaderTest.php
+++ b/tests/php/WpscParseAcceptHeaderTest.php
@@ -75,10 +75,18 @@ class WpscParseAcceptHeaderTest extends TestCase {
 
 	// ── Wildcard behaviour ────────────────────────────────────────────────────
 
-	// Wildcard (*/*) covers text/html when text/html is not explicitly listed.
-	public function test_wildcard_covers_html_when_not_explicit(): void {
-		// */* q=1.0 > json q=0.9 → text/html.
-		$this->assertSame( 'text/html', wpsc_parse_accept_header( '*/*,application/json;q=0.9', $this->json_types ) );
+	// */* wildcard does NOT cover text/html when text/html is not explicit.
+	// A JSON client sending "*/*,application/json;q=0.9" must classify as JSON,
+	// not as text/html, to avoid serving cached HTML to non-browser clients.
+	public function test_wildcard_does_not_cover_html_when_not_explicit(): void {
+		// html absent → effective_html_q = 0.0; json q=0.9 > 0.0 → application/json.
+		$this->assertSame( 'application/json', wpsc_parse_accept_header( '*/*,application/json;q=0.9', $this->json_types ) );
+	}
+
+	// Fediverse regression: "application/json, */*" must NOT serve cached HTML.
+	public function test_fediverse_json_with_wildcard_classifies_as_json(): void {
+		// html absent → effective_html_q = 0.0; json q=1.0 > 0.0 → application/json.
+		$this->assertSame( 'application/json', wpsc_parse_accept_header( 'application/json, */*', $this->json_types ) );
 	}
 
 	// Explicit text/html;q=0.5 wins over */*;q=1.0 — wildcard does not boost an explicit html q.
--- a/tests/php/WpscParseAcceptHeaderTest.php
+++ b/tests/php/WpscParseAcceptHeaderTest.php
@@ -1,0 +1,118 @@
+<?php
+/**
+ * Tests for wpsc_parse_accept_header().
+ *
+ * @package automattic/wp-super-cache
+ */
+
+// wpsc_parse_accept_header() has no WordPress dependencies — include only the
+// containing file. wp_cache_debug() is defined inside wp-cache-phase2.php and
+// returns early when its globals are unset, so no stub is needed. apply_filters()
+// is a WP core function not defined in the file; stub it for completeness.
+if ( ! function_exists( 'apply_filters' ) ) {
+	function apply_filters( $tag, $value ) { // phpcs:ignore Generic.CodeAnalysis.EmptyStatement
+		return $value;
+	}
+}
+
+require_once dirname( __DIR__, 2 ) . '/wp-cache-phase2.php';
+
+use PHPUnit\Framework\TestCase;
+
+/**
+ * @covers wpsc_parse_accept_header
+ */
+class WpscParseAcceptHeaderTest extends TestCase {
+
+	/** Default JSON types mirroring the wpsc_accept_headers default. */
+	private array $json_types = array(
+		'application/json',
+		'application/activity+json',
+		'application/ld+json',
+	);
+
+	// ── RFC 7231 acceptance criteria from issue #1045 ─────────────────────────
+
+	/**
+	 * New Relic Synthetics header: text/html has implicit q=1.0, JSON is
+	 * deprioritised at q=0.9 — must classify as text/html.
+	 */
+	public function test_nr_synthetics_header_classifies_as_html(): void {
+		$accept = 'text/html,application/xhtml+xml,application/json;q=0.9,application/javascript;q=0.9,text/javascript;q=0.9,application/xml;q=0.9,text/plain;q=0.8,*/*;q=0.7';
+		$this->assertSame( 'text/html', wpsc_parse_accept_header( $accept, $this->json_types ) );
+	}
+
+	/** Bare JSON-only Accept: application/json should classify as JSON. */
+	public function test_bare_json_classifies_as_json(): void {
+		$this->assertSame( 'application/json', wpsc_parse_accept_header( 'application/json', $this->json_types ) );
+	}
+
+	/** Tie (both implicit q=1.0) resolves to text/html (safe default). */
+	public function test_tie_resolves_to_html(): void {
+		$this->assertSame( 'text/html', wpsc_parse_accept_header( 'text/html,application/json', $this->json_types ) );
+	}
+
+	/** JSON strictly higher q than text/html must classify as JSON. */
+	public function test_json_higher_q_classifies_as_json(): void {
+		$this->assertSame( 'application/json', wpsc_parse_accept_header( 'application/json,text/html;q=0.9', $this->json_types ) );
+	}
+
+	/** Extended JSON type (application/ld+json) via filter participates in comparison. */
+	public function test_extended_json_type_participates(): void {
+		$this->assertSame( 'application/json', wpsc_parse_accept_header( 'application/ld+json;q=1.0,text/html;q=0.8', $this->json_types ) );
+	}
+
+	/** Malformed q-value (non-numeric) treated as q=1.0, no warning/fatal. */
+	public function test_malformed_q_value_treated_as_default(): void {
+		// application/json;q=bad → treated as q=1.0; text/html is absent → effective html_q from */*=0.7.
+		$this->assertSame( 'application/json', wpsc_parse_accept_header( 'application/json;q=bad,*/*;q=0.7', $this->json_types ) );
+	}
+
+	/** Out-of-range q-value clamped — q=2 treated as 1.0. */
+	public function test_out_of_range_q_clamped(): void {
+		$this->assertSame( 'text/html', wpsc_parse_accept_header( 'text/html;q=2,application/json;q=0.9', $this->json_types ) );
+	}
+
+	// ── Wildcard behaviour ────────────────────────────────────────────────────
+
+	// Wildcard (*/*) covers text/html when text/html is not explicitly listed.
+	public function test_wildcard_covers_html_when_not_explicit(): void {
+		// */* q=1.0 > json q=0.9 → text/html.
+		$this->assertSame( 'text/html', wpsc_parse_accept_header( '*/*,application/json;q=0.9', $this->json_types ) );
+	}
+
+	// Explicit text/html;q=0.5 wins over */*;q=1.0 — wildcard does not boost an explicit html q.
+	public function test_explicit_html_takes_precedence_over_wildcard(): void {
+		// html explicit q=0.5; wildcard q=1.0; json q=0.8 → 0.8 > 0.5 → application/json.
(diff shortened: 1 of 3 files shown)


donnchawp (changes requested):

Thanks for this — the direction is right and the execution is clean: extracting a testable `wpsc_parse_accept_header()` helper, real RFC 7231 q-value parsing, preserved memoization and the `wpsc_accept_headers` filter contract, malformed/out-of-range handling, and no-DB unit tests that fit nicely into the smoke-test tier being set up in #1051. The reported New Relic Synthetics case is fixed correctly.

There's **one required change** before merge: the `*/*` wildcard fallback re-introduces a deliberately-fixed Fediverse regression.

### The problem

This PR was written against an earlier version of the issue brief that still credited `text/html` with the `*/*` quality value. That brief has since been updated to remove it, precisely because of this case. The current code:

```php
} elseif ( null !== $wildcard_q ) {
    $effective_html_q = $wildcard_q;
}
```

Trace a Fediverse fetch — `Accept: application/activity+json, */*`:

- `html_q` = null (no explicit `text/html`)
- `wildcard_q` = 1.0 → `effective_html_q` = 1.0
- `json_q` = 1.0
- `1.0 > 1.0` is false → returns **`text/html`**

The same happens for any JSON client sending `Accept: application/json, */*`. Because the cache only ever *serves* `text/html` (`wp_cache_serve_cache_file()` bails on anything else), classifying these as `text/html` means the client passes the serve-gate and **gets handed the cached HTML page instead of being bypassed to the application** — the exact problem the Accept-header handling was added to prevent. The asymmetry matters here: misclassifying an HTML-preferring request as JSON only costs a cache miss; misclassifying a JSON-preferring request as HTML serves the wrong representation.

### Suggested fix

Use the explicit `text/html` q-value only, defaulting to `0.0` when it's absent — don't fall back to `*/*`:

```php
// Effective html q: explicit text/html only. A */* wildcard must not lift
// text/html above an explicitly requested JSON type, or JSON/Fediverse
// clients sending e.g. "application/json, */*" get served cached HTML.
$effective_html_q = ( null !== $html_q ) ? $html_q : 0.0;
```

…and drop the `$wildcard_q` tracking. This still fixes the New Relic case (explicit `text/html;q=1.0` > `application/json;q=0.9`), and `Accept: */*` on its own still resolves to `text/html` because `json_q` is `0.0` when no JSON type is present.

### Test changes

- `test_wildcard_covers_html_when_not_explicit` asserts the behaviour we need to remove — flip it: `*/*,application/json;q=0.9` should classify as **`application/json`**.
- Add a regression test: `application/json, */*` → `application/json`.
- `test_explicit_html_takes_precedence_over_wildcard` and a `*/*`-alone → `text/html` case both still pass unchanged.

Everything else looks good to go once the wildcard handling is removed. For full background on why JSON/Fediverse traffic is kept out of the cache entirely, see the discussion on #1045.

thisismyurl:
Good catch — the wildcard fallback was definitely wrong for this case. The asymmetry you described is exactly right: misclassifying a JSON client as HTML silently serves the wrong representation, while the reverse only costs a cache miss.

Pushed a follow-up that:
- Drops `$wildcard_q` tracking entirely
- Sets `$effective_html_q = ( null !== $html_q ) ? $html_q : 0.0;` — explicit `text/html` only, no wildcard fallback
- Updates the docblock to document the intentional omission
- Flips `test_wildcard_covers_html_when_not_explicit` → `test_wildcard_does_not_cover_html_when_not_explicit` (asserts `application/json`)
- Adds `test_fediverse_json_with_wildcard_classifies_as_json` as the regression guard for `application/json, */*`

The New Relic Synthetics case still works: explicit `text/html;q=1.0` in the header gives `effective_html_q = 1.0`, which beats `json;q=0.9` as before. `Accept: */*` alone still resolves to `text/html` because both `html_q` and `json_q` are `0.0` and the `>` comparison is false.

donnchawp (commented):

The classification logic is now correct — thanks for turning it around. `effective_html_q` uses the explicit `text/html` q-value only (defaulting to `0.0`), the `*/*` fallback is gone, and `test_fediverse_json_with_wildcard_classifies_as_json` (`application/json, */*` → `application/json`) plus the flipped wildcard test lock in the Fediverse behaviour. No objections to the production change in `wp-cache-phase2.php`.

The two red checks are both test-file hygiene, not logic — all 15 assertions actually pass.

### PHPUnit (PHP 8.2) — doc-comment metadata deprecation

The class-level `/** @covers wpsc_parse_accept_header */` doc-comment triggers a PHPUnit deprecation, and both `phpunit.11.xml.dist` and `phpunit.12.xml.dist` set `failOnPhpunitDeprecation="true"`. The "only 8.2 fails" puzzle is a PHPUnit-version split: on 8.2 composer downgrades to PHPUnit 11, which still reads doc-comment metadata and emits the deprecation → exit 1; on 8.3+ it uses PHPUnit 12, which ignores doc-comment metadata entirely, so no deprecation fires. The run log confirms it: `OK, but there were issues! Tests: 15, Assertions: 15, PHPUnit Deprecations: 1`.

**Fix:** replace the `@covers` doc-comment with the attribute, which works on both 11 and 12 and matches the repo's existing `Jetpack.PHPUnit.Attributes` sniff:

```php
use PHPUnit\Framework\Attributes\CoversFunction;

#[CoversFunction( 'wpsc_parse_accept_header' )]
class WpscParseAcceptHeaderTest extends TestCase {
```

### PHPCS — 7 errors, all in `tests/php/WpscParseAcceptHeaderTest.php`

1. **"A file should either contain function declarations or OO structure declarations, but not both."** The `apply_filters()` stub function collides with the test class. `wpsc_parse_accept_header()` (the function under test) doesn't call `apply_filters`, so the stub is probably unnecessary; if requiring `wp-cache-phase2.php` turns out to need it, move it into `tests/php/bootstrap.php`. Either way the test file should contain only the class.
2. Missing `@var` tag on the `$json_types` property docblock.
3. The four `//` method comments (the wildcard-behaviour notes) need to be `/** ... */` docblocks.
4. One doc-comment short description starts lowercase ("application/activity+json classified…") — capitalize it.

### Minor

The comment in `test_malformed_q_value_treated_as_default` still says "effective html_q from `*/*`=0.7" — leftover from the old wildcard logic. The assertion is correct, but the reasoning should read "html absent → effective_html_q = 0.0; json q=1.0 > 0.0 → application/json."

Once the `@covers` becomes an attribute and the PHPCS items are cleared, this should go green. Logic LGTM.

donnchawp:
Thanks @thisismyurl for the PR. Almost there. :)

thisismyurl:
Pushed a follow-up addressing both CI failures.

**PHPUnit deprecation (PHP 8.2)** — replaced `/** @covers wpsc_parse_accept_header */` with `#[CoversFunction( 'wpsc_parse_accept_header' )]` + `use PHPUnit\Framework\Attributes\CoversFunction`. The attribute works on PHPUnit 11 and 12; the doc-comment form only triggered the deprecation on 11 (the PHP 8.2 path).

**PHPCS — 7 errors** — all in the test file:
- Moved the `apply_filters()` stub to `tests/php/bootstrap.php` (where it belongs as test infrastructure); fixed the bootstrap `@package` tag while there.
- Added `@var string[]` to the `$json_types` property docblock.
- Converted the four `//` wildcard-behaviour method comments to `/** ... */` docblocks.
- Updated the lowercase short description: "application/activity+json classified as JSON" → "Classifies application/activity+json as JSON".
- Fixed the stale inline comment in `test_malformed_q_value_treated_as_default` — removed the `*/*=0.7` reasoning and updated to `effective_html_q = 0.0`.

donnchawp:
Thank you @thisismyurl ! It's good to go!

PR #1067, opened 2026-06-22
Introduce a Config module to own the config-file write path (#1062)

### LINKED ISSUE #1062: Refactor: introduce a Config module to own the config-file write path
> *This issue was generated by AI.*

Concrete, tiny-commit implementation plan for **#1047 item #1** (a `Config` module that owns the on-disk config file). Sequenced **after the `wp-cache.php` file move (#1061)** and building on the PHPUnit test net (#1051). Scope was deliberately narrowed in a planning interview to the **write path only**: globals stay exactly as they are; reads are untouched.

## Problem Statement

WP Super Cache's configuration is ~63 loose PHP variables (≈100 counting `$wp_cache_pages[...]` sub-keys) defined in `wp-cache-config.php`. The write path is the dangerous part:

- Writes go through two mechanisms: `wp_cache_setting()` (53 call sites) and **direct calls to `wp_cache_replace_line()` (64 call sites)** — a ~100-line function that rewrites a live PHP file with a per-line regex.
- More than half of all writes bypass the `wp_cache_setting()` wrapper and call the raw regex rewriter directly, each passing its own hand-built regex and replacement line.
- There is no single owner of the file's format. The effective "schema" is the union of 117 regex/line pairs scattered across the codebase.
- The file-rewriting logic has **no test coverage**, so any change to it is unverifiable.

This makes adding a setting a three-part chore (a new global, a new write call with a bespoke regex, a new read site) and makes the rewriter impossible to refactor safely.

## Solution

Introduce a single **`Config` module** that owns reading and writing the config file and is the only code that knows the file's on-disk format. Every write funnels through it.

Crucially, this is an **additive abstraction, not a replacement of the globals**:

- The globals stay populated exactly as today (`Config::set()` updates `$GLOBALS[$field]` just like `wp_cache_setting()` already does). All 193 read sites (`global $foo;`, `$GLOBALS['foo']`, and the settings-map's dynamic lookup) keep working unchanged.
- `wp_cache_setting()` and `wp_cache_replace_line()` **remain as global functions** — they are part of the public surface third-party cache plugins may call — but become thin delegations to the module.
- The dangerous regex-rewriting logic moves into one place, gets a small enough surface to test thoroughly, and the 64 direct callers migrate to a typed `Config::set($key, $value)` API.

The module follows the established `src/` convention (namespace `Automattic\WPSC`, `class-*.php` filename, explicit `require_once` at runtime — there is **no runtime autoloader**; the Composer classmap is dev/test-only). It is loaded explicitly from `wp-cache.php` for the admin write path and lazily (guarded by `class_exists`) from the two phase2 write shims so the debug-log write path keeps working when only the engine is loaded.

### Phase A — Lock down current write behaviour first

1. **Characterization tests for the existing write path.** Against a temporary config file, assert the exact resulting file contents after `wp_cache_setting()` for each value type (numeric, boolean, array/object via `var_export`, string) and after representative direct `wp_cache_replace_line()` calls covering its three branches: replace an existing line, the "setting unchanged" early return, and the "not found → insert after the defines/assignments" append path. Also assert `$GLOBALS[$field]` is updated. These tests encode today's behaviour byte-for-byte and become the regression guard for everything that follows.

### Phase B — Introduce the owner (no caller changes yet)

2. **Add the `Config` class.** Create the class under `src/` following the existing convention, moving the file read/format/write logic into methods: a writer that performs the current atomic `tempnam` → write → `rename` → `chmod` → `opcache_invalidate` sequence, and a value formatter that reproduces the numeric/bool/array/string line formatting. Wire runtime loading: explicit `require_once` in `wp-cache.php`, and lazy `require_once` (guarded by `class_exists`) in the phase2 shims. No existing caller changes; the class is loadable but unused. Add direct unit tests mirroring the Phase A characterization tests against the class methods.

3. **Delegate `wp_cache_replace_line()` to the module.** Replace the function body with a call into the `Config` writer; keep the signature identical. Run the Phase A characterization suite — output must be byte-identical. This routes all 64 direct callers through the owner without touching any of them.

4. **Delegate `wp_cache_setting()` to the module.** Replace its body with a `Config::set()` call (which formats the value, updates `$GLOBALS`, and writes). Run the suite — identical behaviour. After this commit there is exactly one owner of the file format and all 117 write sites flow through it via the two shims.

### Phase C — Migrate direct callers to the typed API

Each commit converts one cluster of direct `wp_cache_replace_line( '^ *\$foo', "\$foo = ...;", $file )` calls into `Config::set( 'foo', $value )`, verified against the test suite and the e2e settings specs. Grouped by the (now-split, post-#1061) file/cluster so each commit is small and revertable:

5. **Lifecycle writes** — activation/deactivation/enable/disable toggles (e.g. enabling/disabling caching).
6. **Settings-form writes** — the rejected/accepted-list, tracking-parameter, and debug-settings updaters.
7. **Admin-UI / manager writes** — the `wp_cache_manager_updates` settings-save path.
8. **Preload writes** — preload-related settings persisted from the preload UI/cron.
9. **Remaining scattered writes** — htaccess/mod-rewrite flags, cron-check, version stamps, and any stragglers.

Calls that are not expressible as a simple key/value `set()` (e.g. removing a line, conditional `define` rewrites, or multi-line edits) stay on the raw `wp_cache_replace_line()` function — which still works, now delegating — and are listed explicitly so the exceptions are visible.

10. **Optionally retire `wp_cache_setting()` internal usage.** Convert the 53 internal `wp_cache_setting()` calls to `Config::set()` for consistency. The function itself stays as a public back-compat alias. Cosmetic; can be dropped if not worth the churn.

### Phase D — Document

11. **Document the module and the contract.** Update `CONTEXT.md` and add an ADR stating: all config writes go through `Config`; the globals/`$GLOBALS` remain a permanent, supported read API; reads are intentionally **not** migrated; and the config file stays a flat pre-WordPress PHP file. Note the locking decision and its rationale.

## Decision Document

- **Write path only.** This pass centralises and tames *writes*. The ~193 read sites (`global $foo;`, `$GLOBALS['foo']`, settings-map dynamic reads) are untouched.
- **Globals are permanent back-compat.** `Config` always populates `$GLOBALS[$field]` on write, exactly as `wp_cache_setting()` does today. Existing installs' `wp-cache-config.php` files and third-party cache plugins (e.g. bundled `plugins/multisite.php`, `wptouch.php`, `domain-mapping.php`, and the external `wp-super-cache-plugins` ecosystem) that read these globals keep working indefinitely. The globals are treated as a supported read API, not deprecated.
- **The config file stays a flat PHP file loaded pre-WordPress.** The drop-in (`wp-cache-phase1.php`) `@include`s it before WP boots, with no database available. It cannot move to `wp_options`. `Config` reads/writes that same file format.
- **`wp_cache_setting()` and `wp_cache_replace_line()` remain global functions** as delegating shims, because they are callable from third-party code. They are not removed.
- **No new locking or concurrency handling.** Per maintainer input: writes happen only in wp-admin, effectively always a single human changing one setting, and the path has been stable for years. The existing atomic `tempnam`+`rename` is preserved; file locking is explicitly out of scope.
- **The module builds on the existing settings map.** `rest/class.wp-super-cache-settings-map.php` already maps setting → `{global, option, get/set method}`; the `Config` key set aligns with it rather than inventing a parallel schema.
- **No runtime autoloader.** Confirmed in-code ("WPSC doesn't use an autoloader or composer"). `Config` is loaded by explicit `require_once` at runtime (mirroring `plugins/jetpack.php`'s loading of the device-detection class) and is classmap-autoloaded only under test.
- **Depends on #1061.** Phase C clusters map onto the post-split `inc/` files, so the file move lands first.

## Testing Decisions

- **A good test here asserts the on-disk result and the populated global, not internals.** The binding contract is "the config file ends up in this exact state and `$GLOBALS[$field]` holds this value." Tests write to a temp file and assert file contents + global state for representative value types and rewriter branches.
- **The write path is the strongest unit-test target in the plugin** — it is nearly pure (a file in, a file out) once isolated. Characterization tests are written *before* the module exists (Phase A) and re-run unchanged against the module (Phase B) to prove byte-identical behaviour.
- **Modules tested:** the `Config` writer and value formatter, plus the `wp_cache_setting()` / `wp_cache_replace_line()` delegations. The per-caller migrations in Phase C are covered by the same suite plus the existing e2e settings specs (`tests/e2e/specs/settings/`).
- **Tiering (per #1051):** these config write tests touch WP runtime helpers (e.g. `wp_rand`, transients via `set_transient`), so they run in the **local integration tier** via the Makefile target against the Dockerized database — not CI. CI stays lint + a no-database smoke suite. (If the writer's WP dependencies are trivially stubbable, individual cases may drop to the smoke tier, but assume integration by default.)
- **Prior art:** the integration test harness established in #1051 (run locally via the Makefile target) and the e2e settings tests as the behavioural backstop.

## Out of Scope

- **Read migration.** Converting `global $foo;` / `$GLOBALS['foo']` reads to `Config::get()` is a separate, much larger and riskier effort touching the engine and third-party compat. Not included.
- **Removing or deprecating the globals.** They remain a permanent read API.
- **Moving config off the flat PHP file** (e.g. into `wp_options` or a serialized store). Impossible given the pre-WordPress drop-in load.
- **File locking / concurrency safety.** Explicitly declined (single-human admin writes).
- **The REST read pattern.** `rest/class.wp-super-cache-rest-get-settings.php` and `-get-status.php` re-`include()` the config file to refresh globals on read; that is the read path and is left as-is (the module could own a reload later).
- **Any engine/serving logic** in `wp-cache-phase2.php` beyond the two write shims.

## Further Notes

- This is item #1 of the #1047 umbrella and is the recommended foundation for the later `CacheKey` (#2) and `Preload` state machine (#5) work, since those will want a clean config surface.
- Highest-risk commits are the Phase C migrations where a direct `wp_cache_replace_line()` call did something subtler than a key/value set; the explicit "exceptions stay on the raw API" rule contains that.
- Sequencing rationale: the maintainer chose to do the file move (#1061) before this, even though #1047 suggested config-first — fine, because the conservative write-only scope here doesn't require chasing reads, and the post-split files make the Phase C clusters cleaner.
- **Constraint for the future `Config::get()` read API (out of scope here, but recorded so it isn't lost):** a `get()` will be added later, but the globals must remain because third-party code depends on them — so `get()` and the globals must be a *single* source of truth, not two. `Config::get($key)` must be a typed read-through accessor over `$GLOBALS` (the canonical store, populated from the flat file pre-WordPress), **not** a value served from a private copy that `Config` parsed separately. Otherwise a third-party plugin that writes a global directly would be invisible to `get()` (and vice versa) and the two would drift. This is symmetric with the write side, which already keeps `$GLOBALS` authoritative. Casting/typing belongs inside `get()`, since the globals hold whatever raw PHP values the config file's `include` produced.

donnchawp:
> *This PR was generated by AI.*

Implements **#1062** — a `Config` module that owns the on-disk `wp-cache-config.php` write path. Conservative, behaviour-preserving refactor: the write path is centralised and made testable while runtime behaviour, the globals read API, and the public function surface stay exactly as they were.

Closes #1062.

## What changed

- **New owner.** `Automattic\WPSC\Config` (`src/config/class-config.php`) is the single owner of the config-file format: `set()` (updates `$GLOBALS[$field]`, formats, writes), `format_value()` (per-type encoding), and `write_line()` (the atomic `tempnam`→`rename`→`chmod`→`opcache_invalidate` rewriter). `write_line()` is a verbatim transplant of the old `wp_cache_replace_line()` body.
- **Public functions become delegating shims.** `wp_cache_setting()` and `wp_cache_replace_line()` keep identical signatures and remain in `wp-cache-phase2.php` (third-party cache plugins call them); they lazy-load the class via a `class_exists` + `file_exists` guard so the engine-only/debug-log path keeps working.
- **Caller migration.** 39 direct `wp_cache_replace_line()` callers across `inc/` that are simple `$field = value;` writes were migrated to `Config::set()`. 50 internal `wp_cache_setting()` callers were converted to `Config::set()` for consistency (phase2's engine-path debug-log calls deliberately stay on the guarded shim).
- **Documented exceptions.** Writes whose on-disk format isn't reproducible by `Config::format_value()` (define rewrites, line removals, `var_export` `cache_path`, imploded `wp_cache_mobile_groups`, numeric-but-quoted `cache_time_interval`/`cache_max_time`, `cache_page_secret`, `$wp_cache_pages[…]` sub-keys, the `___`-encoded `cache_rejected_user_agent`, the unanchored `sem_id` regex) stay on the raw function — catalogued in ADR 0002.
- **Scope held tight.** No read site migrated; no `Config::get()`; no new locking; no runtime autoloader (explicit `require_once`; classmap is dev/test-only).

## Out of scope (recorded in ADR 0002)

Read migration (`global $foo` / `$GLOBALS`), removing/deprecating the globals (a permanent read API), moving config off the flat pre-WordPress PHP file, and file locking.

## Tests

- **New:** `tests/php/integration/ConfigWritePathTest.php` (characterizes the write path byte-for-byte), `ConfigClassTest.php` (the class API), plus 3 schedule-cluster cases added to `SettingsFormUpdatersTest.php` that pin the quoted-vs-unquoted `cache_time_interval`/`cache_max_time` distinction.
- **Docs:** ADR `docs/adr/0002-config-module-write-path.md` + a `CONTEXT.md` entry.

## Test plan

- [x] `composer test-php` — smoke suite (no DB): 34/34
- [x] `make test-integration` — WP + DB integration: 62/62 (incl. `WpCacheSettingBackCompatTest` locking the public shim contract)
- [x] `composer test-e2e` — Docker/Jest e2e: 6 suites / 28 tests
- [x] `composer lint` — changed-lines PHPCS: exit 0

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -47,6 +47,18 @@ defined in time for their hooks.
   follow-up, separate from the relocation. Newly authored `inc/` files remain
   fully linted.
 
+## Config write path
+
+`src/config/class-config.php` (`Automattic\WPSC\Config`) is the single owner of
+the on-disk `wp-cache-config.php` format. All config-file writes flow through
+`Config::set()` / `Config::write_line()`. The global functions `wp_cache_setting()`
+and `wp_cache_replace_line()` in `wp-cache-phase2.php` are delegating shims kept
+for third-party compatibility; they lazy-load the class so engine-only loads
+remain safe. `$GLOBALS` variables remain the canonical, supported read API —
+reads are not migrated. See `docs/adr/0002-config-module-write-path.md` for the
+full decision, constraints, and the catalog of writes that deliberately bypass
+`Config::set()`.
+
 ## Tests
 
 Two tiers (see `tests/php/README.md`):
--- a/docs/adr/0002-config-module-write-path.md
+++ b/docs/adr/0002-config-module-write-path.md
@@ -101,7 +101,8 @@ simple `$field = value;` key/value assignments. These are intentional, not misse
 | `cache_rejected_user_agent` | The `___`↔space round-trip leaves the in-memory array holding `___` placeholders at write time while the written string has spaces restored, so `Config::set()` (which formats the array) would write different bytes |
 | `sem_id` write in `inc/lifecycle.php` | Uses a bare unanchored regex (`'sem_id'`), not the `'^ *\$sem_id'` anchored form `Config::set()` generates; migrating would change which lines the regex matches |
 | Debug-log writes in `wp-cache-phase2.php` | Stay on `wp_cache_setting()` for engine-only load safety (the lazy-load guard); the shim handles this correctly |
-| `plugins/*` and `rest/*` direct calls | Third-party-facing callers are out of scope for this refactor |
+| `plugins/*` direct calls | Third-party-facing plugin integrations shipped with the plugin; out of scope for this refactor |
+| `set_ossdlcdn` in `rest/class.wp-super-cache-rest-update-settings.php` | Raw integer-in-string write (`"\$ossdlcdn = $ossdlcdn;"`); matches `format_value`'s unquoted-int output but was left on the raw call. The handler's other `wp_cache_setting()` calls (already routed through the same `is_numeric` logic) were migrated to `Config::set()` |
 
 ## Consequences
 
--- a/docs/adr/0002-config-module-write-path.md
+++ b/docs/adr/0002-config-module-write-path.md
@@ -1,0 +1,118 @@
+# 2. Config module write-path
+
+Date: 2026-06-22
+
+## Status
+
+Accepted
+
+## Context
+
+After splitting `wp-cache.php` into focused `inc/` files (ADR 0001), config-file
+writes were still scattered across those files as direct calls to the
+`wp_cache_setting()` and `wp_cache_replace_line()` global functions. Neither
+function was tested, their on-disk format rules lived only in `wp_cache_setting()`
+at line ~1400 of `wp-cache-phase2.php`, and there was no clear owner of the
+`wp-cache-config.php` file format. Issue #1062 called for consolidating the write
+path into a testable class.
+
+The constraint the config file is operating under is fundamental: `wp-cache-phase1.php`
+`@include`s it before WordPress boots — no database, no `wp_options` — so it must
+remain a flat, pre-WordPress PHP file that sets global variables.
+
+## Decision
+
+### `Automattic\WPSC\Config` owns the config-file write path
+
+`src/config/class-config.php` (`Automattic\WPSC\Config`) is the single owner of
+the on-disk `wp-cache-config.php` format. All config-file writes are expected to
+flow through it. Its public surface:
+
+- **`Config::set( $field, $value, $file = null )`** — updates `$GLOBALS[$field]`,
+  formats the value, and writes the line. The `$GLOBALS` update is unconditional:
+  the runtime state and the file are always kept in sync.
+- **`Config::format_value( $value )`** — encodes the PHP literal written to the
+  file: numeric → unquoted; bool → `true`/`false`; array/object → whitespace-
+  collapsed `var_export`; else → single-quoted string.
+- **`Config::write_line( $old, $new, $file )`** — the atomic line rewriter: reads
+  the file, replaces the line matching `$old` (a bare regex), writes to a
+  `tempnam`-generated temp file, `rename`s it into place, `chmod`s it to match
+  the original permissions, and calls `opcache_invalidate`.
+
+### `wp_cache_setting()` and `wp_cache_replace_line()` remain as delegating shims
+
+Both functions stay in `wp-cache-phase2.php` with their original signatures. They
+are part of the surface that third-party cache plugins call, so they are not
+removed. Each shim lazy-loads the class via:
+
+```php
+if ( ! class_exists( 'Automattic\WPSC\Config' ) ) {
+    if ( file_exists( WPCACHEHOME . 'src/config/class-config.php' ) ) {
+        require_once WPCACHEHOME . 'src/config/class-config.php';
+    }
+}
+```
+
+The `file_exists` guard means the debug-log write path keeps working when only the
+engine (`wp-cache-phase2.php`) is loaded and the admin tree is absent.
+
+### Globals remain the canonical read API
+
+`Config::set()` always writes to `$GLOBALS[$field]`. Reads (`global $foo;` /
+`$GLOBALS['foo']` / settings-map dynamic lookups) are intentionally not migrated —
+that is a separate, larger effort and out of scope here. Any future `Config::get()`
+must be a typed read-through over `$GLOBALS` (the single source of truth, populated
+from the flat file before WordPress boots), not a value served from a private copy
+the class parsed separately. If `Config` maintained its own parsed copy, a
+third-party plugin writing a global directly would be invisible to `get()` and the
+two would drift — the same symmetry as the write side, which keeps `$GLOBALS`
+authoritative.
+
+### No new locking or concurrency
+
+The existing atomic `tempnam`+`rename` sequence is preserved unchanged. File
+locking is explicitly out of scope: admin config writes are effectively
+single-human, and the rename is already atomic at the OS level.
+
+### No runtime autoloader
+
+The class is loaded by explicit `require_once` — from `wp-cache.php` on the admin
+path and lazily from the two phase2 shims. The Composer classmap (`composer.json`
+`autoload.classmap`) is dev/test-only and not loaded at runtime.
+
+### Exception catalog: writes that deliberately stay on `wp_cache_replace_line()`
+
+Most direct `wp_cache_replace_line()` callers in the `inc/` files were migrated to
+`Config::set()`. The following writes remain on the raw function because their
+on-disk format is not reproducible by `Config::format_value()` or they are not
+simple `$field = value;` key/value assignments. These are intentional, not misses:
+
+| Exception | Reason |
+|-----------|--------|
+| `define( ... )` rewrites (`WP_CACHE`, `WPCACHEHOME`, `WPLOCKDOWN`) | Not `$var =` assignments; some target the site `wp-config.php`, not the config file |
+| Line removals (`$new = ''`) | Removal is not a value write; `Config::set()` has no delete path |
+| `cache_path` | Written via `var_export()`, which escapes backslashes and quotes differently from `format_value`'s string branch |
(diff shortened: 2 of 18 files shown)


donnchawp:
> *This comment was generated by AI.*

## Reviewer testing checklist

This is a behaviour-preserving refactor, so the pass condition for most checks is: **save a settings group, then `diff` the generated `wp-cache-config.php` against a pre-PR copy — byte-identical (modulo the value you changed).** Do this on a real install with caching + `WP_CACHE` enabled.

### Highest value — the one path CI can't cover
- [ ] **Engine-only / drop-in lazy-load.** The `Config` class lazy-loads inside the `wp_cache_setting()` / `wp_cache_replace_line()` shims; every automated test loads the full plugin, so this branch is never exercised by CI. Trigger the **debug-log write path** (Settings → Debug: set/change the debug log file or debug username) and confirm the config file updates with no fatal. This runs the shim with the class *not* pre-loaded.
- [ ] **Unusual install layout.** If you can, test where `WPCACHEHOME` resolves oddly (symlinked plugin dir, or the `realpath( WPCACHEHOME ) != realpath( __DIR__ )` branch in `wp-cache.php`) — confirm the shim finds `src/config/class-config.php`. (If `src/` is ever missing *and* the class isn't loaded, the shim now fatals where the old code returned `false` — only on a broken/partial install.)

### Settings clusters with quoting subtleties (save → diff config file)
- [ ] **GC / preload schedule** (`cache_schedule_type`, `cache_time_interval`, `cache_max_time`): confirm `cache_time_interval` stays **quoted** and `cache_max_time` stays unquoted-in-settings-forms / quoted-in-admin-ui. (Pinned by `SettingsFormUpdatersTest`, but the most likely place a real divergence would show.)
- [ ] **Rejected/accepted URIs, rejected user-agents, tracking parameters, rejected cookies** — array writes; the user-agent one has the `___`/space round-trip.
- [ ] **Advanced toggles** (304, mfunc, mobile, compression, not-logged-in, late-init, …) — lowest risk; covered by e2e.

### Alternative entry points (route through the same shims)
- [ ] **REST settings save** (the React admin UI, if used) — exercises the shim from a different path than the classic forms.
- [ ] **Bundled plugins** that call `wp_cache_replace_line()` directly (not migrated, now route through the shim): WPtouch, Jetpack mobile, domain mapping, bad-behaviour, awaiting-moderation. Confirm their settings still persist.

### Two more
- [ ] **opcache on:** confirm config changes take effect without a manual flush (the writer's `opcache_invalidate()` is preserved verbatim).
- [ ] **`$GLOBALS` sync (intentional):** migrated raw-rewriter sites now *also* write `$GLOBALS[$field]`. Verified inert; if a settings page reflects a *fresher* value than before, that's the reason — a fix, not a regression.

### Automated (already green at branch head)
- [x] `composer test-php` — 34/34 · `make test-integration` — 62/62 (incl. `WpCacheSettingBackCompatTest`) · `composer test-e2e` — 6 suites / 28 · `composer lint` — exit 0

PR #1080, opened 2026-08-10
Encoded from unicode parts of URI should be uppercased in the both cases

IvanZhuck:
When we have a post slug which contains unicode characters like this:

_ұlytau-oblysy-ortasha-ajlyқ-zhalaқy-kөlemi-bojynsha-elimizde-kөsh-bastady_

the plugin creates a directory which contains encoded parts instead of unicode:

_%D2%B1lytau-oblysy-ortasha-ajly%D2%9B-zhala%D2%9By-k%D3%A9lemi-bojynsha-elimizde-k%D3%A9sh-bastady_

the encoded parts are uppercased. It work when we read cache, but when we save the post, the plugin tries to remove the cache directory. It attempts to locate the directory with lowercased post slug but the directory isn't exist, so the cache couldn't be refreshed when the post saved.

Please review the fix for the issue.

### Release Notes
- Fix: clear the cache correctly when saving a post whose slug contains non-ASCII characters.

### DIFF
--- a/tests/php/integration/SupercacheDirCaseTest.php
+++ b/tests/php/integration/SupercacheDirCaseTest.php
@@ -1,0 +1,119 @@
+<?php
+/**
+ * Case-normalisation tests for get_current_url_supercache_dir().
+ *
+ * WordPress stores a non-ASCII post slug with LOWERCASE percent escapes:
+ * utf8_uri_encode() builds them with dechex(), and sanitize_title_with_dashes()
+ * lowercases the result afterwards. The supercache directory on disk, however,
+ * is written with those escapes UPPERCASED, because browsers send them that way
+ * and the mod_rewrite rules do a literal file test on the request path.
+ *
+ * Only the $post_id = 0 branch ever creates a supercache directory; the
+ * $post_id != 0 branch is used exclusively by the deletion paths. So whatever
+ * the $post_id = 0 branch produces is the shape the other branch has to match,
+ * or cache invalidation on post save looks for a directory that is not there.
+ * See PR #1080.
+ *
+ * Integration tier: needs get_permalink() and the rewrite rules, so these run
+ * under the real WordPress runtime.
+ *
+ * @package automattic/wp-super-cache
+ */
+class SupercacheDirCaseTest extends WP_UnitTestCase {
+
+	/**
+	 * Kazakh title from the original bug report, which sanitises to a slug full
+	 * of percent escapes.
+	 */
+	const NON_ASCII_TITLE = 'Ұlytau oblysy ortasha ajlyқ zhalaқy kөlemi';
+
+	public function set_up() {
+		parent::set_up();
+
+		$GLOBALS['cache_path']           = '/tmp/wpsc-supercache-dir-test/';
+		$GLOBALS['WPSC_HTTP_HOST']       = 'example.org';
+		$GLOBALS['wp_cache_home_path']   = '/';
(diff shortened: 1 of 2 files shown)


donnchawp:
Thanks Ivan for that. I remember this came up a couple of times over the years but I couldn't find a fix. I found the same issue elsewhere, so I'll update your PR with those fixes too.

donnchawp:

Thanks again Ivan. I pushed two commits on top of yours.

**199112c** - your change moved the escape-uppercasing so both branches run it, but the `$post_id` branch still never lowercased the path. On a subdirectory install at `/Blog/` it built `/Blog/%D2%B1.../` while the directory on disk is `/blog/%D2%B1.../`, so saving a post still missed the cache directory there. Only the `$post_id = 0` branch ever creates a supercache directory, so I hoisted its `strtolower()` out to sit beside the escape-uppercasing and let both branches share it. Lowercase first, then uppercase the escapes.

The same commit adds `tests/php/integration/SupercacheDirCaseTest.php`, covering both halves. All three tests fail against trunk as it was before this PR. They only run under `make test-integration` and not in CI, since they need `get_permalink()` and the rewrite rules.

**6f432a1** - three test-only cleanups:

- The class docblock said the mod_rewrite rules do a literal file test on the request path, and that's not why the escapes are uppercase. In .htaccess context Apache matches `RewriteRule` against the decoded path, so `$1` expands to raw UTF-8 and the `-f` test never finds a percent-encoded directory. The real reason is simpler: `sanitize_title_with_dashes()` lowercases the escapes it stores in `post_name`, and a URL encoder emits them uppercase, so the same path arrives spelled two ways. That one's mine, I wrote the wrong reason down in 199112c.
- `set_up()` left `$wp_cache_home_path` at `/Blog/` for every test class that ran after it. It now snapshots the globals it overwrites and puts them back in `tear_down()`.
- `get_current_url_supercache_dir()` memoises into a function static keyed by `$post_id` with no way to reset it, so the first caller with key 0 wins for the whole run. There's now an assertion at the top of the key 0 test, so if a future test gets there first it fails saying that, rather than as an unexplained directory mismatch.

`make test-integration` is green (40 tests, 95 assertions), and so is `composer test-php`.

One thing I checked and want written down: the write path output is byte identical before and after all of this, so nothing already on disk is stranded and no upgrade routine is needed. Only the delete path moves, and only from wrong to right.

I opened #1081 for the same lowercase/uppercase mismatch in `wpsc_delete_url_cache()`, which is what the taxonomy, post type and author archive deletions go through. A post in a category with a non-ASCII slug still leaves a stale archive. Not blocking this PR.

donnchawp:

Two things since the last comment, neither of them anything for you to do, Ivan.

**The red PHPUnit checks are ours, not yours.** The 8.2 and 8.3 legs are failing because our `composer.lock` can't be installed on those versions any more, and the fallback `composer update` trips over a new advisory against PHP_CodeSniffer, CVE-2026-67434. I re-ran trunk's last green build, same commit, nothing changed in between, and it fails the same way, so it's rot rather than anything in this patch. Tracked in #1082. PHPCS, E2E, and PHPUnit on 8.4 and 8.5 are all green.

**#1081 grew a fourth call site.** `wp_cron_preload_cache()` at inc/preload.php:229 builds the same path by hand with no normalisation. It's the worst of them, because it deletes and then immediately re-fetches, so when the delete misses the fetch is served from the cache file still sitting there and nothing regenerates. A taxonomy archive with a non-ASCII slug is never refreshed by preload at all. Still out of scope here.

